PR #12, opened 2021-12-17
[RND-203] Query parameter protections

stephenfuqua:


stephenfuqua:
We seem to have abandoned this. The ticket is still open, so we'll come back to it eventually. Leaving the draft PR here.

stephenfuqua:
This has been "rebased" (actually recreated) due to the significant restructuring of the code that took place after the original commit.

Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/QueryOpensearch.ts
```diff
@@ -80,15 +91,14 @@ export async function queryDocuments(request: QueryRequest, client: Client): Pro
       query += ` WHERE ${whereClause}`;
     }
 
+    query += ' ORDER BY _doc';
+
```
stephenfuqua, line 95: OpenSearch is strict that `ORDER BY` must come after `LIMIT ... OFFSET ...`.

Meadowlark-js/packages/meadowlark-core/src/validation/DocumentValidator.ts
```diff
@@ -34,6 +34,21 @@ export type DocumentValidationResult = {
...
+  errorBody?: string;
+};
+
+async function getMatchingMetaEdModel(pathComponents: PathComponents): Promise<TopLevelEntity | undefined> {
+  const lowerResourceName = decapitalize(pathComponents.endpointName);
+  const modelNpmPackage = modelPackageFor(pathComponents.version);
+  const { metaEd } = await loadMetaEdState(modelNpmPackage);
+
+  return getMetaEdModelForResourceName(lowerResourceName, metaEd, pathComponents.namespace);
+}
```
stephenfuqua, line 50: Extracted from below for reuse.

Meadowlark-js/packages/meadowlark-core/src/validation/PaginationValidator.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-use-before-define */ // function declarations are hoisted properly in ES6
```
stephenfuqua, line 6: @bradbanister as you know I like private local functions. What do you think about tweaking the eslint file to allow this pattern, assuming it is possible?

My understanding is that function expressions are no longer hoisted in ES6, so they wouldn't be appropriate. But regular function declarations are still hoisted and thus perfectly safe.

Meadowlark-js/tests/http/local.opensearch.http
```diff
@@ -0,0 +1,93 @@
...
+authorization: bearer {{authToken1}}
+
+### 400: invalid query term
+GET http://localhost:3000/local/v3.3b/ed-fi/students?asdfasdfa=another
+authorization: bearer {{authToken1}}
+
+### 400: malformed query term
+GET http://localhost:3000/local/v3.3b/ed-fi/students?birth%27City=another&another=another
+authorization: bearer {{authToken1}}
+
```
stephenfuqua, line 75: I ran the tests below on the old code. Was "fun" to see the request getting back more results than we would have liked. The final `#` completely comments out the ownership check.

Meadowlark-js/packages/meadowlark-core/src/validation/PaginationValidator.ts
bradbanister: I actually like that eslint rule because it makes code browsing easier. You always know which direction to scroll when referring to a called function. Could you relocate them to the beginning of the function? Alternatively, I'd stick with the eslint disable, but do it per-line rather than whole file.

bradbanister (approved):
A few suggestions

Meadowlark-js/packages/meadowlark-core/src/validation/PaginationValidator.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export function validatePaginationParameters(parameters: PaginationParameters): string | undefined {
+  // Acceptable for neither to be defined
+  if (parameters.limit == null && parameters.offset == null) {
+    return undefined;
+  }
+
+  const limit: string[] = [];
+  const offset: string[] = [];
+
```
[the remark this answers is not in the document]
stephenfuqua: I see where you are coming from, but if I do that, then I'll have to rename the properties when creating the HTTP response message, which should be of the form similar to the ODS/API:

```
{
    "message": "The request is invalid.",
    "modelState": {
        "request.Limit": [
            "The value 'a' is not valid for Limit."
        ]
    }
}
```

Trying to match the ODS/API error messages is a new idea that came up from the SIG. We'll need to change other validation processes in a future ticket. I did not try to match precisely at this point, because "request.Limit" feels too much like an artifact of .NET.

Meadowlark-js/packages/meadowlark-core/src/validation/PaginationValidator.ts
stephenfuqua: Thanks to hoisting, that too will work 😄 . In C#, the private functions must be at the end if they are accessing variables from outside their scope. But in JS/TS this is perfectly fine.

Personally I have a small preference for them at the end of the function, so that I can quickly scan the "heart" of the function without needing to look past the private function definitions. But that is unusual, and I have definitely heard some complaints / confusion about that. So let's try out putting them at the top.

Meadowlark-js/packages/meadowlark-core/src/validation/PaginationValidator.ts
```diff
@@ -0,0 +1,53 @@
...
+  validateLimit();
+  validateOffset();
+
+  if (limit.length > 0 || offset.length > 0) {
+    return createInvalidRequestResponse({ limit, offset });
+  }
```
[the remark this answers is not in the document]
stephenfuqua: I decided to make this change, except as `limitErrors` and `offsetErrors`

PR #17, opened 2022-04-07
[RND-209] add mongodb container to docker-compose

bradbanister:


Meadowlark-js/docker-meadowlark/docker-compose.yml
```diff
@@ -37,9 +39,20 @@ services:
       DISABLE_SECURITY_DASHBOARDS_PLUGIN: true # disables security dashboards plugin in OpenSearch Dashboards
     networks:
       - opensearch-net
+    restart: unless-stopped
+
+  mongodb-node1:
+    image : mongo:4.0.28
```
bradbanister, line 45: Chose MongoDB v4.0 because it is the latest version AWS DocumentDB is compatible with. From initial research, MongoDB 4.0 has a sufficient feature set meet our needs.

stephenfuqua: I'm surprised they're compatible with 4.0, I had in back of my mind that this was the first version after license change.

Unfortunately, [Azure is only up to 3.6 if I'm reading correctly](https://docs.microsoft.com/en-us/azure/cosmos-db/mongodb/feature-support-36#:~:text=is%20three%20years.-,Protocol%20Support,Cosmos%20DB's%20API%20for%20MongoDB.).

stephenfuqua: Nope, just realized that they do have more than this, I just happened to land on an older page first :-).
https://docs.microsoft.com/en-us/azure/cosmos-db/mongodb/feature-support-42

bradbanister: Yeah, Cosmos DB is ahead of DocumentDB right now. https://docs.aws.amazon.com/documentdb/latest/developerguide/compatibility.html

bradbanister: Interestingly, the DocumentDB defaults to 3.6 compatibility. You have to ask for 4.0 support. We'd figure it out pretty quick though because 4.0 is the first version to support transactions, and I'm pretty sure we're going to need them just like with DynamoDB.

stephenfuqua: 4.0 sounds good. I'll try this out as soon as possible.

Meadowlark-js/docker-meadowlark/readme.md
```diff
@@ -46,9 +47,11 @@ At a command prompt, running in this directory:
 | Operation | Command |
 | -- | -- |
 | start containers | `docker compose up -d` |
+| view running containers | `docker container ls` |
```
stephenfuqua, line 50: Gives the same information:

```suggestion
| view running containers | `docker ps` |
```

stephenfuqua (approved):

PR #18, opened 2022-04-11
[RND-210] Proof of concept - implement POST and GET by Id on MongoDB without reference validation

bradbanister:


bradbanister:
This is a PoC of POST and GET by Id on MongoDB using some of the RND-210 design elements.

I implemented a simple `MongoEntityRepository` with the same functions as in `DynamoEntityRepository`. Only `createEntity` and `getEntityById` are implemented, and there's no reference validation. Worked pretty well though:
image

Meadowlark-js/src/repository/MongoEntityRepository.ts
```diff
@@ -0,0 +1,145 @@
...
+// See the LICENSE and NOTICES files in the project root for more information.
+
+import { Collection, Db, MongoClient, WithId } from 'mongodb';
+import { ForeignKeyItem } from '../model/ForeignKeyItem';
+import { EntityInfo } from '../model/EntityInfo';
+import { Security } from '../model/Security';
+import { ValidationOptions } from '../model/ValidationOptions';
+import { DeleteResult, GetResult, OwnershipResult, PutResult } from './BaseDynamoRepository';
+
+interface Entity {
```
stephenfuqua, line 13: I wonder if we could use this exact same structure on PostgreSQL, which has a JSON data type? Keep the single-table design so that we don't have to generate schema-code.

But that doesn't help us get foreign keys, which would be the primary reason to consider such an idea. Secondary reason would be to give people SQL instead of NoSQL.

Just thinking "out loud" here.

Meadowlark-js/src/repository/MongoEntityRepository.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+async function naiveGetEntities(): Promise<Collection<Entity>> {
+  const client = new MongoClient('mongodb://localhost:27017');
+
+  await client.connect();
+  const db: Db = client.db('meadowlark');
+  const entities: Collection<Entity> = db.collection('entities');
+
+  // Note this will trigger a time-consuming index build if the indexes do not already exist.
```
stephenfuqua, line 44: I wonder if we can pre-create these indexes in a full implementation?

Maybe just uploading (and then retrieving) descriptors would be sufficient?

stephenfuqua:
The spike is helpful. I want to keep this branch without merging, so I'll close the PR. Going to rename the branch to `mongo-poc` so that it is easy to find without having to remember the ticket number.

PR #20, opened 2022-04-20
[RND-227] Refactor to plugin architecture

bradbanister:


Meadowlark-js/packages/meadowlark-core/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@edfi/meadowlark-core",
+  "version": "0.2.0-dev.0",
+  "description": "Meadowlark core functionality",
+  "license": "SEE LICENSE IN LICENSE.md",
```
stephenfuqua, line 5: Would it makes sense to change this for Meadowlark?
```suggestion
  "license": "Apache-2.0",
```

Meadowlark-js/packages/meadowlark-core/src/handler/CrudHandler.ts
```diff
@@ -107,7 +98,7 @@ export async function create(event: APIGatewayProxyEvent, context: Context): Pro
     }
 
     const resourceId = documentIdForEntityInfo(entityInfo);
-    const { result, failureMessage } = await createEntity(
+    const { result, failureMessage } = await backendPlugin().createEntity(
```
stephenfuqua, line 101: I had to stop and think about this and the dependency injection pattern. Seems like a variant on the Service Lookup pattern.

Meadowlark-js/packages/meadowlark-core/src/handler/CrudHandler.ts
```diff
@@ -6,16 +6,6 @@
 /* eslint-disable-next-line import/no-unresolved */
```
stephenfuqua, line 6: Some of the functions in this module are very lengthy. Personally, I usually prefer < 15 lines in a function. Looking at the details, I'm not sure that's realistic. I think there's also a bunch of refactoring we can do when we get to the point of teasing apart the front end and the business logic or service layer.

Just mentioning this for discussion.

Meadowlark-js/packages/meadowlark-core/src/handler/NaturalKeyExtractor.ts
```diff
@@ -122,8 +123,8 @@ export function extractNaturalKey(entity: TopLevelEntity, body: object): Natural
  * TODO: Bring this behavior into extractNaturalKey, correctly managing for natural key component ordering.
  */
 export function deriveAssignableFrom(entity: TopLevelEntity, naturalKey: string): AssignableInfo | null {
-  const { assignableTo }: { assignableTo: TopLevelEntity | null } = (entity.data
-    .meadowlark as EntityMeadowlarkData).apiMapping;
+  const { assignableTo }: { assignableTo: TopLevelEntity | null } = (entity.data.meadowlark as EntityMeadowlarkData)
+    .apiMapping;
```
stephenfuqua, line 127: I'm having trouble understanding what this `{ .. } : { ... } = ...` is actually doing. What is the data type on the left?

Meadowlark-js/packages/meadowlark-core/src/handler/NaturalKeyExtractor.ts
```diff
@@ -88,9 +88,10 @@ function naturalKeysFrom(
  * extracts the natural key information from the JSON body. Also extracts security information, if any.
  */
 export function extractNaturalKey(entity: TopLevelEntity, body: object): NaturalKeyWithSecurity {
-  const naturalKeysWithSecurity: NaturalKeyWithSecurity[] = (entity.data
-    .meadowlark as EntityMeadowlarkData).apiMapping.identityReferenceComponents.flatMap(
-    (identityReferenceComponent: ReferenceComponent) => naturalKeysFrom(identityReferenceComponent, body, entity, []),
+  const naturalKeysWithSecurity: NaturalKeyWithSecurity[] = (
+    entity.data.meadowlark as EntityMeadowlarkData
+  ).apiMapping.identityReferenceComponents.flatMap((identityReferenceComponent: ReferenceComponent) =>
+    naturalKeysFrom(identityReferenceComponent, body, entity, []),
```
stephenfuqua, line 94: That's a lot of code to read in this one statement. Makes me wonder if some of this "work" should be done inside another class? I don't know the object model here anymore, so this question might not make any sense.

Meadowlark-js/packages/meadowlark-core/src/plugin/PluginLoader.ts
```diff
@@ -0,0 +1,18 @@
...
+import { MeadowlarkBackendPlugin } from './backend/MeadowlarkBackendPlugin';
+import { NoMeadowlarkBackendPlugin } from './backend/NoMeadowlarkBackendPlugin';
+
+dotenv.config();
+
+const loadedBackendPlugin = process.env.BACKEND_PLUGIN_MODULE
+  ? require(`../packages/${process.env.BACKEND_PLUGIN_MODULE}/index`).initializeBackendPlugin()
+  : NoMeadowlarkBackendPlugin;
+
+export function backendPlugin(): MeadowlarkBackendPlugin {
```
stephenfuqua, line 16: When I first read code above that calls `backendPlugin()` I thought it might be building something with every call. I wonder if it would be a little easier to read  with a small rename: `getBackend()` perhaps? Something to think about; I'm not fixated on it.

Meadowlark-js/packages/meadowlark-core/src/plugin/backend/SearchResult.ts
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+/** TODO: this needs to be unified with GetResult */
```
stephenfuqua, line 6: When will we do that? Do we have a natural ticket for it?

Meadowlark-js/packages/meadowlark-plugin-dynamodb-opensearch/src/types/AwsSdkLibDynamoDb.ts
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: Apache-2.0
```
stephenfuqua, line 1: Is this old code that just moved to a new file?

stephenfuqua (commented):
Some of my comments are about old code. All comments are up for conversation.

Meadowlark-js/packages/meadowlark-core/package.json
bradbanister: Done

Meadowlark-js/packages/meadowlark-core/src/handler/CrudHandler.ts
bradbanister: Agree, they are longer than I'd like. As you say, I think some of this will end up spilt between front end and meadowlark-core.

Meadowlark-js/packages/meadowlark-core/src/handler/NaturalKeyExtractor.ts
bradbanister: This is the somewhat odd Typescript syntax for annotating an object destructuring assignment. entity.data.meadowlark.apiMapping is of type ApiEntityMapping.  ApiEntityMapping has a property assignableTo, and  assignableTo is of type TopLevelEntity | null.

Three options here: 1) Leave as is. 2) Drop it completely - not strictly necessary as TypeScript knows the type. 3) Introduce a local type alias like `type NullableTopLevelEntity = TopLevelEntity | null` and use that to replace `{ assignableTo: TopLevelEntity | null }`.

Meadowlark-js/packages/meadowlark-core/src/handler/NaturalKeyExtractor.ts
bradbanister: This statement looks more complicated than it is. Everything before the `flatMap` is just referencing the `identityReferenceComponents` of the `entity`, and `flatMap` is calling a single function `naturalKeysFrom`. The type annotations make it look somewhat verbose.

Meadowlark-js/packages/meadowlark-core/src/plugin/PluginLoader.ts
bradbanister: Sounds good. `backendPlugin` was originally a `const` but became a function to make mocking out easy.

Meadowlark-js/packages/meadowlark-core/src/plugin/backend/SearchResult.ts
bradbanister: I think this will happen as we make the backend plugin interface generic.

Meadowlark-js/packages/meadowlark-plugin-dynamodb-opensearch/src/types/AwsSdkLibDynamoDb.ts
bradbanister: It's a temporary duplication due to an existing leaky abstraction. The AWS type `PutItemInputAttributeMap` is in the meadowlark model object ForeignKeyItem. meadowlark-plugin-dynamodb-opensearch is the correct place for this file. I reduced the duplicate in meadowlark-core to just define `PutItemInputAttributeMap`. Later removal will be obvious when the time comes, because it currently requires meadowlark-core to depend on AWS libraries.

Meadowlark-js/packages/meadowlark-core/src/handler/NaturalKeyExtractor.ts
stephenfuqua: Option 3 sounds good. This sort of thing is powerful I'm sure, but confusing for those of us who don't live and breathe TypeScript.

bradbanister:
Note that this PR merges into mongo-poc to avoid cluttering this PR with those changes. Once this is approved and merged, we should follow up with a PR to merge mongo-poc into main.

bradbanister:
In addition to factoring out backend plugins, this PR has the very beginning of factoring out the frontend AWS service. It also has a large portion of the MetaEd-specific code factored out into a true MetaEd plugin for Meadowlark.

stephenfuqua (commented):
Code looks good... haven't tested yet, will try to do so soon.

stephenfuqua:
I'm getting some unit tests failures on this branch, after checking out the branch, running `yarn install & yarn build`.

```
Test Suites: 3 failed, 21 passed, 24 total
Tests:       6 failed, 198 passed, 204 total
Snapshots:   27 passed, 27 total
```

bradbanister:
The integration tests were intermittently failing because the dynamodb instance wasn't ready. Moved to separate "yarn test:integration" script.

stephenfuqua:
Even with the latest commits, and after `rm -r node_modules & yarn install & yarn test:unit` - I still get four failed tests that "exceeded timeout of 5000 ms for a hook". 

* Two failures point to: GetResolver.test.ts:90
* and two point to: Create.test.ts:468

Time permitting I will try increasing the jest timeout for those tests.

stephenfuqua:
I added a commit that increases timeout on the failing tests to 6 seconds and then everything passed. I was able to switch to the between DynamoDB and MongoDB plugins with a simple `.env` change, and interact with the API successfully.

stephenfuqua (approved):

PR #24, opened 2022-04-28
[RND-225] Mongo 3 node replica set

bradbanister:


Meadowlark-js/docker-meadowlark/docker-compose.yml
```diff
@@ -41,18 +41,77 @@ services:
...
+    container_name: mongo2
+    image: mongo:4.0.28
+    expose:
+      - 27018
+    ports:
+      - 27018:27018
+    networks:
+      - mongo-net
+    restart: always
+    entrypoint: [ "/usr/bin/mongod", "--bind_ip_all", "--replSet", "rs0", "--journal", "--dbpath", "/data/db", "--enableMajorityReadConcern", "false", "--port", "27018" ]
```
stephenfuqua, line 85: Just curious - why run each on a different port?

Oh, because all three nodes are exposed outside of the Docker network. Wonder if that's really advisable? Would be nice to have a single point of entry.

Meadowlark-js/packages/meadowlark-plugin-mongodb/src/MongoEntityRepository.ts
```diff
@@ -42,7 +42,7 @@ interface Entity {
 }
 
 async function naiveGetEntities(): Promise<Collection<Entity>> {
-  const client = new MongoClient('mongodb://localhost:27017');
+  const client = new MongoClient('mongodb://localhost:27017,localhost:27018,localhost:27019');
```
stephenfuqua, line 45: Should this be

```suggestion
  const client = new MongoClient('mongodb://localhost:27017,localhost:27018,localhost:27019/?replicaSet=rs0');
```

Meadowlark-js/docker-meadowlark/docker-compose.yml
bradbanister: Ended up being necessary for the hosts file trick to work

Meadowlark-js/packages/meadowlark-plugin-mongodb/src/MongoEntityRepository.ts
bradbanister: Doesn't seem to be necessary

bradbanister:
Ready for final review, needs addition of `127.0.0.1 mongo1 mongo2 mongo3` to hosts file.

Meadowlark-js/docker-meadowlark/scripts/mongo-setup.sh
```diff
@@ -0,0 +1,48 @@
...
+            "_id": 1,
+            "host": "mongo2:27018",
+            "priority": 0
+        },
+        {
+            "_id": 2,
+            "host": "mongo3:27019",
+            "priority": 0
+        }
+    ],settings: {chainingAllowed: true}
```
stephenfuqua, line 41: FYI, this configuration will not elect a different primary if `mongo1` goes down. I tried to get that to happen - documentation implies that it should work, if all have the same priority and all have a vote (which is not how the above is configured). Still doesn't work. When I admin'ed MongoDB cluster some years back, my third node was set as an arbiter and it would choose the primary. Of course, what happens if the arbiter is down?

Anyway, after investigating this for too many minutes, I'm going to back away - we're not trying to get a production setup here. Just something that will work for our "POC".

stephenfuqua (approved):

PR #37, opened 2022-06-02
[RND-279] Protect against revealing the OpenSearch password in logs

stephenfuqua:


stephenfuqua:
## Before
_Note_ that the username and password are set to the same value here.


image


## After

image

stevenarnold-ed-fi (approved):

PR #39, opened 2022-06-06
RND-240 - Get and Upsert w/o reference validation

stevenarnold-ed-fi:


Meadowlark-js/backends/meadowlark-postgres-backend/src/security/SecurityResponse.ts
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: Thinking down the road, once we get PostgreSQL running, we should look for commonality with the MongoDB backend to extract to a library, maybe named "meadowlark-backend-support"

Meadowlark-js/backends/meadowlark-postgres-backend/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@edfi/meadowlark-postgres-backend",
```
bradbanister, line 2: How about "meadowlark-postgresql-backend"?

Meadowlark-js/backends/meadowlark-postgres-backend/src/repository/Read.ts
```diff
@@ -0,0 +1,43 @@
+// // SPDX-License-Identifier: Apache-2.0
+// // Licensed to the Ed-Fi Alliance under one or more agreements.
+// // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// // See the LICENSE and NOTICES files in the project root for more information.
+
+import { Client, Result } from 'pg';
+import { GetResult, GetRequest, DocumentIdentity } from '@edfi/meadowlark-core';
+
+// @ts-ignore
```
bradbanister, line 9: What doesn't TS like about this?

Meadowlark-js/backends/meadowlark-postgres-backend/src/repository/Db.ts
```diff
@@ -0,0 +1,130 @@
...
+  database: 'meadowlark',
+};
+
+export async function checkExistsAndCreateTables(client: Client) {
+  try {
+    await client.query(`CREATE SCHEMA IF NOT EXISTS meadowlark`);
+
+    const documentsTableSql =
+      'CREATE TABLE IF NOT EXISTS meadowlark.documents(' +
+      '_pk bigserial PRIMARY KEY,' +
```
bradbanister, line 26: Do we need this? Our document id is unique and identifying, so could it be the pk?

bradbanister: Otherwise, we'd also need an index on id

Meadowlark-js/backends/meadowlark-postgres-backend/src/repository/Db.ts
```diff
@@ -0,0 +1,130 @@
...
+      'id character varying NOT NULL,' +
+      'document_identity JSONB NOT NULL,' +
```
bradbanister, line 28: Interesting. I probably would have done as a string but this might be better.

Meadowlark-js/backends/meadowlark-postgres-backend/src/security/SecurityResponse.ts
stevenarnold-ed-fi: agreed

Meadowlark-js/backends/meadowlark-postgres-backend/package.json
stevenarnold-ed-fi: updated

Meadowlark-js/backends/meadowlark-postgres-backend/src/repository/Read.ts
stevenarnold-ed-fi: in it's current state, it's fine and I forgot to remove the ignore, not remembering what caused me to put that there in the first place

Meadowlark-js/backends/meadowlark-postgres-backend/src/repository/Db.ts
stephenfuqua: I haven't yet looked to see if there is a definitive answer, and I think it verges on "religious war." Personally I prefer to have a synthetic numeric primary key in a relational database because of advantages with storage size and sorting behavior. Yes, will therefore need an index on our natural key id in that case.

Meadowlark-js/backends/meadowlark-postgres-backend/src/repository/Db.ts
```diff
@@ -0,0 +1,130 @@
...
+      'is_descriptor boolean NOT NULL,' +
+      'validated boolean NOT NULL,' +
+      'edfi_doc JSONB NOT NULL);';
+
+    await client.query(documentsTableSql);
+
+    const referencesTableSql =
+      'CREATE TABLE IF NOT EXISTS meadowlark.references(' +
+      '_pk bigserial PRIMARY KEY,' +
+      'references_to character varying NOT NULL,' +
```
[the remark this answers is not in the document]
stephenfuqua: Please no composite primary keys 😨

Meadowlark-js/backends/meadowlark-postgres-backend/src/repository/Db.ts
bradbanister: In a typical relational model I prefer surrogate keys. The complexity of the ODS/API is a good example of why.  I'll bet you could greatly simplify the ODS/API code if the DB used surrogate keys instead of cascading composite key columns on to sub-tables. I know MetaEd generation would be significantly simpler. We certainly wouldn't have needed the "merge" keyword, with all its issues and complexities.

The reason I'm thinking we don't need surrogate keys is that this isn't a relational model - at least not yet. We don't need to join the references table back to the documents table. That would over complicate things and be needlessly slower. All we need the reference table for is delete reference checking. That's just a simple query of the table for any row with a matching references_to entry. Any results means the delete should be failed.

It's entirely possible that we'll want a relational model between the documents table and some future security tables, but that's something we can figure out later. Until then, it would be nice to keep things simple.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Db.ts
```diff
@@ -0,0 +1,130 @@
...
+      'resource_version character varying NOT NULL,' +
+      'is_descriptor boolean NOT NULL,' +
+      'validated boolean NOT NULL,' +
+      'edfi_doc JSONB NOT NULL);';
+
+    await client.query(documentsTableSql);
+
+    const referencesTableSql =
+      'CREATE TABLE IF NOT EXISTS meadowlark.references(' +
+      '_pk bigserial PRIMARY KEY,' +
```
stephenfuqua, line 40:
```suggestion
      'id bigserial PRIMARY KEY,' +
```

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Db.ts
```diff
@@ -0,0 +1,130 @@
...
+      'references_to character varying NOT NULL,' +
+      'references_from character varying NOT NULL);';
```
stephenfuqua, line 42:
```suggestion
      'parent_document_id character varying NOT NULL,' +
      'referenced_document_id character varying NOT NULL);';
```

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -0,0 +1,74 @@
+// // SPDX-License-Identifier: Apache-2.0
+// // Licensed to the Ed-Fi Alliance under one or more agreements.
+// // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// // See the LICENSE and NOTICES files in the project root for more information.
+
+import { DeleteResult, Logger, DeleteRequest } from '@edfi/meadowlark-core';
+import { Client } from 'pg';
+import { deleteDocumentByIdSql } from './QueryHelper';
+
+// This function should only be used for testing, it will clear out all data from both documents and references tables
```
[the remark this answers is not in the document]
stevenarnold-ed-fi: done

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -0,0 +1,74 @@
...
+export async function deleteAll(client: Client): Promise<DeleteResult> {
+  const deleteResult: DeleteResult = { response: 'UNKNOWN_FAILURE' };
+
+  try {
+    await client.query('BEGIN');
+    await client.query('DELETE FROM meadowlark.documents;');
+    await client.query('DELETE FROM meadowlark.references;');
+    await client.query('COMMIT');
```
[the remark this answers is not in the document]
stevenarnold-ed-fi: I think it's fine to have separate calls for each of these[[this is an example](https://github.com/brianc/node-postgres/issues/1252#issuecomment-293899088)](url) from the library creator, where he does something similar. But since this is a simple case with all the SQL together, I have no issue changing it to a single SQL statement

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -0,0 +1,74 @@
...
+    deleteResult.response = 'DELETE_SUCCESS';
+  } catch (e) {
+    Logger.error('postgres.repository.Upsert.upsertDocument', 'DELETE_ALL', e);
+    await client.query('ROLLBACK');
```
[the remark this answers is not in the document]
stevenarnold-ed-fi: Postgres does not automatically rollback. From examples from the library's creator (see the link above) it's standard to do a rollback in the catch block

Meadowlark-js/backends/meadowlark-postgresql-backend/test/integration/Read.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  afterAll(async () => {
+    await deleteAll(client);
+    await resetSharedClient();
+  });
+
+  it('should return the document', async () => {
+    expect(getResult.response).toBe('GET_SUCCESS');
+    expect(getResult.document.edfiDoc.inserted).toBe('yes');
```
bradbanister, line 111: This isn't right. Your code is returning something like a `DocumentInfo` as `document` on the `GetResult`. However, the `document` _is_ the `edfiDoc` as defined on a `DocumentInfo`. The correct expectation should be `expect(getResult.document.inserted).toBe('yes');`

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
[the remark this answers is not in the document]
stephenfuqua: @bradbanister any preference here? I don't actually have a strong feeling, if it does work as expected.

Meadowlark-js/backends/meadowlark-postgresql-backend/test/integration/Read.test.ts
stevenarnold-ed-fi: Ah, got it. I added the values we stored in the database to the document, including the JSON edfi_doc but didn't realize that I needed to iterate over the JSON and add those as well.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Db.ts
```diff
@@ -0,0 +1,111 @@
...
+  return dbPool.connect();
+}
+
+/**
+ * Return the shared client
+ */
+export async function getSharedClient(): Promise<Client> {
+  if (dbPool == null) {
+    const client = await createConnectionPoolAndReturnClient();
+    await checkExistsAndCreateTables(client);
```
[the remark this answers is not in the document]
stevenarnold-ed-fi: @stephenfuqua what about moving the checkExistsAndCreateTables call into the end of the createConnectionPoolAndReturnClient call? This would mean it would only check/create tables if the meadowlark database doesn't exist rather than every startup, leaving the happy path as 

1. Create Pool
2. Connect to Pool
3. Return connected client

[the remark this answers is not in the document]
stephenfuqua: I thought about that too, but then what happens if someone pre-creates the empty database?

Hence let's just leave as is and defer until later, after we get feedback from the community about deployment approaches.

bradbanister (approved):
Looks good!

PR #47, opened 2022-06-20
[RND-243] Postgres Support with Reference Validation

stevenarnold-ed-fi:


Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -32,41 +41,45 @@ export async function deleteDocumentById(
...
+      if (referencesResult.rows[0].exists) {
+        Logger.debug(
+          `postgres.repository.Delete.deleteDocumentById: Deleting document id ${id} failed due to existing references`,
+          traceId,
+        );
+
+        // Get the DocumentIdentities of up to five referring documents for failure message purposes
+        const referringDocuments = await client.query(getReferencedByDocumentSql(id));
+        const failures: string[] = referringDocuments.rows.map(
+          (document) => `Resource ${document.resource_name} with identity '${JSON.stringify(document.document_identity)}'`,
```
stephenfuqua, line 63: We may need to be more subtle than this - in a future ticket though.

In other words, in what situations would it be unwise to reveal the document identity for documents that referred to the one being deleted?

1. Referring document is owned by a different client.
2. Referring document has PII in the identity.

On the second point, that might not be an issue if we can resolve the first point. After all, if the client "owns" the referring document, then why should we be concerned about an HTTP response containing identity information? In theory they send the body with that PII in the first place!

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Upsert.ts
```diff
@@ -4,26 +4,43 @@
...
   let isInsert: boolean;
 
+  const outRefs = documentInfo.documentReferences.map((dr: DocumentReference) => documentIdForDocumentReference(dr));
+
   try {
-    recordExistsResult = await client.query(getRecordExistsSql(id));
+    recordExistsResult = await client.query(getCheckDocumentExistsSql(id));
 
-    isInsert = !recordExistsResult.rowCount || recordExistsResult.rowCount === 0;
+    isInsert = !recordExistsResult.rows[0].exists;
```
stephenfuqua, line 38: There's 0.0% chance that `rows == null`?

stevenarnold-ed-fi: I believe so, since the query is now a sub select it should return true or false no matter what, but will research a bit to verify

stephenfuqua: I would assume that the PostgreSQL driver might react in some strange way and return nothing. Because if we don't assume that, and it does happen, then we end up with a nasty error that is harder to debug.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/QueryHelper.ts
```diff
@@ -88,12 +153,49 @@ export function getDocumentInsertOrUpdateSql(
...
 }
 
+// SQL for Deletes
+
+/**
+ * Returns the SQL query for deleting a document from the database
+ * @param documentId the document to delete from the documents table
+ * @returns SQL query string to delete the document
+ */
+export function deleteDocumentByIdSql(documentId: string): string {
```
bradbanister, line 163: Most of your SQL builder functions have a "get" prefix, but this one doesn't. I actually like how this reads better, as the "get" prefixes feel redundant. It's already clear from the "Sql" suffixes that they return SQL.

stevenarnold-ed-fi: I've gone back and forth on the naming of these functions, they didn't originally have the get or sql has part of the name, and thus looked like they would do the work and not just return the SQL. then I added the 'get' to resolve and when I wasn't satisfied, I added the sql suffix. I'm still not super happy but see your point with not having the 'get' so I'll remove them

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Upsert.ts
```diff
@@ -4,26 +4,43 @@
...
-  // let outRefs;
-  let documentSql: string;
+  let documentUpsertSql: string;
   let isInsert: boolean;
 
+  const outRefs = documentInfo.documentReferences.map((dr: DocumentReference) => documentIdForDocumentReference(dr));
+
   try {
-    recordExistsResult = await client.query(getRecordExistsSql(id));
+    recordExistsResult = await client.query(getCheckDocumentExistsSql(id));
```
bradbanister, line 36: I don't think this works. Your existence check here happens outside the transaction.

More broadly though, PostgreSQL has an upsert syntax that would simplify this. Here's an example of the syntax: https://towardsdatascience.com/postgresql-how-to-upsert-safely-easily-and-fast-246040514933

bradbanister: Say the document already exists. This query returns true, and you are ready for an update. Then a different request deletes the document before you enter your transaction. In your transaction, you choose the "UPDATE" SQL, and error because you are updating something that isn't there.

This does bring up another issue not handled here. Say we bring the check into the transaction (we should), but otherwise assume the same sequence. The transaction itself will fail, which is fine but it needs to be retried (with some sort of backoff). The MongoDB driver handles retries for you in this situation, but I'm assuming we'll need to handle it ourselves here.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Upsert.ts
stevenarnold-ed-fi: Discussed with Stephen re: hardening error detection overall for PostgeSQL will move forward as is and revisit in RND-299

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Upsert.ts
stevenarnold-ed-fi: I'm not sure why I originally had the check outside of the transaction, but I went ahead and moved it into the transaction (and a single try/catch)

PR #51, opened 2022-06-28
RND-278 - Unified System Tests

stevenarnold-ed-fi:


Meadowlark-js/backends/meadowlark-postgresql-backend/test/integration/Delete.test.ts
```diff
@@ -76,7 +76,8 @@ describe('given the delete of a non-existent document', () => {
 
   afterAll(async () => {
     await deleteAll(client);
-    await resetSharedClient();
+    client.release();
```
stevenarnold-ed-fi, line 79: Realized that the Postgres deleteAll function has a client.release in it. Would rather not have a release in that function so updated the integration tests to release after calling deleteAll

Meadowlark-js/tests/system/crud/Delete.test.ts
```diff
@@ -3,37 +3,46 @@
...
 
 describe('given a DELETE of a non-existent school', () => {
-  let client: MongoClient;
+  let client;
   let deleteResult: FrontendResponse;
 
   beforeAll(async () => {
-    client = (await getNewClient()) as MongoClient;
-    await getCollection(client).deleteMany({});
+    client = await backendToTest.TestingSetup();
```
stevenarnold-ed-fi, line 34: This is the other oddity I mentioned Monday, With Mongo, if I try to end the pool in the setup function and then recreate in the teardown it causes connection issues, if I leave the client w/o passing it to the teardown it hangs Jest. The only way I could get it to work is to pass the client to the teardown and close it there. This code technically works and meets the tickets requirements for not having backend specific code in the tests, but it is a little weird.

Meadowlark-js/tests/system/crud/SystemTestSetup.ts
```diff
@@ -95,7 +95,7 @@ export function academicWeekBodyClient1(): FrontendRequest {
 }
 
 export function configureEnvironmentForMongoSystemTests(): void {
-  process.env.DOCUMENT_STORE_PLUGIN = MONGO_DOCUMENT_STORE_PLUGIN;
+  // process.env.DOCUMENT_STORE_PLUGIN = MONGO_DOCUMENT_STORE_PLUGIN;
```
stevenarnold-ed-fi, line 98: I don't intend to leave this function as is, just haven't fixed it yet.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/index.ts
```diff
@@ -11,6 +11,7 @@ import {
   updateDocumentById,
   securityMiddleware,
 } from './BackendFacade';
+import { TestingTeardown, TestingSetup, TearDownAndReleasePool } from './repository/TestHelper';
```
stevenarnold-ed-fi, line 14: Currently have two teardown functions for Postgres issues - With system tests if I end and null the Postgres connection pool in between describe blocks the tests fail. This appears to be a case where Jest is not running sequentially even with the runInBand option. In the integration tests we end and null the pool between tests with no issue.

stevenarnold-ed-fi: I did test that the postgres integration tests would function properly w/o ending the pool in-between blocks, so maybe that's the update to make. But I find it weird the two different sets of tests are acting differently

stephenfuqua (approved):

PR #62, opened 2022-07-22
RND-261 - Full access authorization mode

stevenarnold-ed-fi:


Meadowlark-js/packages/meadowlark-core/src/security/JwtValidator.ts
```diff
@@ -41,6 +41,17 @@ export function validateJwt(authorizationHeader: string | undefined): JwtValidat
     };
   }
 
+  if (jwtStatus.authorizationStrategy === '') {
```
stevenarnold-ed-fi, line 44: This overall works, but there is an issue (maybe?) - when you have an invalid token (for example the token doesn't parse at all) the error returned is this one, even if this is not the sole reason for the bad token.

stephenfuqua: I'm going to drop a link to this comment in the tech docs page where I'll be working on design plans for full oauth2 support. Agreed this is not great, and that now is not the time to improve it.

Meadowlark-js/packages/meadowlark-core/src/security/JwtStatus.ts
```diff
@@ -15,6 +15,8 @@ export interface JwtStatus {
   subject: string | null;
   issuedAt: number;
   expiresAt: number;
+  roles: string[];
+  authorizationStrategy: string;
```
stephenfuqua, line 19: Why type this as a `string` instead of `AuthorizationStrategy`?

stephenfuqua (commented):
I may have overlooked it... but if not... could you add at least one test, in `local.ownership-tests.http`, to show that a full-access token (roles: ['host']) _can_ access a resource created by one of the vendor tokens?

Meadowlark-js/packages/meadowlark-core/src/security/JwtStatus.ts
stevenarnold-ed-fi: As is, AuthoruzationStrategy doesn't have an UNKNOWN (or similar) type of strategy, so nothing to set it to if it can't be parsed. I wasn't sure if adding to AuthoruzationStrategy was the correct way to go, but if you don't see a problem with it, it would probably be cleaner

stephenfuqua: If it can't be parsed to a known strategy, then isn't that an error that should generate a 500 response?

stevenarnold-ed-fi: I was missing a couple of things in the code. We do have a UNDEFINED AuthorizationStrategy, so I was able to update this to not use a string (and not need the empty string above). But your question makes me wonder - what is the apropriate error response to throw if there is role defined and thus the token is bad. I think a 500 would be incorrect here and it should be a 400, 401 or 403 (currently it's a 401)

Meadowlark-js/packages/meadowlark-core/src/security/JwtValidator.ts
stevenarnold-ed-fi: AuthorizationStrategy has an UNDEFINED (not sure how I missed that)  so the highlighted line looks like `jwtStatus.authorizationStrategy === 'UNDEFINED'`

stevenarnold-ed-fi:
> I may have overlooked it... but if not... could you add at least one test, in `local.ownership-tests.http`, to show that a full-access token (roles: ['host']) _can_ access a resource created by one of the vendor tokens?

Yes - Added a test for POST and DELETE of a student using FULL_ACCESS  auth

Meadowlark-js/packages/meadowlark-core/src/security/JwtStatus.ts
stephenfuqua: Here's why I thought of 500: right now, that means that our code that _generated_ the token did something wrong. But since this is at a latter point, and that code to generate the token could be 3rd party in the future, I agree with you that 500 is actually not appropriate. Its simply a bad token and therefore unauthorized. Ideally, our code would include a message stating that the token is missing role information. Would that be easy to do in this ticket? Or better to put into a stand-alone ticket?

stevenarnold-ed-fi: image
This is the output when roles can not be parsed

stevenarnold-ed-fi: Interestingly, because it would be rare to have an invalid token and yet have roles, this error is now basically the default error if the token is bad

Meadowlark-js/tests/http/local.ownership-tests.http
```diff
@@ -2,8 +2,8 @@
...
-@auth_header_2=Authorization: bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJlZC1maS1tZWFkb3dsYXJrIiwiYXVkIjoibWVhZG93bGFyayIsInN1YiI6InNtYWxsLXRvd24tc2lzIiwianRpIjoiMGQwZWZmYTctYWExZi00N2E3LTk4NzktNDdhOGJkOWMzMWJiIiwiaWF0IjoxNjM2NTYyMDc5LCJleHAiOjM4NDU1NDg4ODF9.2QGyvx9flXdM9wjEHX_VPmeacCI3nm8WcV-T5AWxeWo
+@auth_header_1=Authorization: bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJlZC1maS1tZWFkb3dsYXJrIiwiYXVkIjoibWVhZG93bGFyayIsInJvbGVzIjpbInZlbmRvciJdLCJzdWIiOiJzdXBlci1ncmVhdC1TSVMiLCJqdGkiOiIyNGMyMDIyMC1mMjg1LTRlNjMtYThmNC1iMmRkYzM5MzAwMDQiLCJpYXQiOjE2NTg1MjM0MDgsImV4cCI6MjY1ODUyNzAwOH0.S2C44HoFMr5ZYSUkbZoAV2pz7Z1Xi40yynH17GOBVoY
+@auth_header_2=Authorization: bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJlZC1maS1tZWFkb3dsYXJrIiwiYXVkIjoibWVhZG93bGFyayIsInJvbGVzIjpbInZlbmRvciJdLCJzdWIiOiJzdXBlci1ncmVhdC1TSVMiLCJqdGkiOiI4MjVkNDRiOC1hODExLTRjZmYtYmVmOC1jYzhhNjYxNjliZTAiLCJpYXQiOjE2NTg1MjQyMzUsImV4cCI6MjY1ODUyNzgzNX0.DR9-ixkAs-EUIsoRjYYGY-NLal2o8wrDYYLcKFfb3_I
```
stephenfuqua, line 6: @stevenarnold-ed-fi these two tokens are for the same SIS ("super-great-sis"), which invalidates the tests in this file. For example, the post on line 23 no longer fails.

I generated a new long-lived token for the second set of credentials:

```
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJlZC1maS1tZWFkb3dsYXJrIiwiYXVkIjoibWVhZG93bGFyayIsInJvbGVzIjpbInZlbmRvciJdLCJzdWIiOiJzdXBlci1ncmVhdC1TSVMiLCJqdGkiOiIyMDJhY2Y0My0xZDNkLTQ5YzMtYTZiMy1lNzIxZmQ5ZjE3NGYiLCJpYXQiOjE2NTg3NzQ3NzUsImV4cCI6MjUyNzQ1OTIwMH0.X4gV3613Xq3oDwGh7RSQfbZVPhc13NuYUpibp1_rOxc
```

Even with that in place, the tests are not failing properly. For example, the first token creates an `EducationContents` resource, and the second token should not be able to access it.

I recommend also adding a third set of credentials in `HardcodedCredentials.ts`, just for host provider. Maybe...

```
export const client3: HardcodedCredential = {
  key: 'host_key',
  secret: 'host_secret',
  vendor: 'hosting-provider',
};
```

Then automatically grant this set of credentials, when received, the correct Role of "Host".

stephenfuqua (changes requested):

Meadowlark-js/tests/http/local.ownership-tests.http
```diff
@@ -2,8 +2,8 @@
...
-@auth_header_2=Authorization: bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJlZC1maS1tZWFkb3dsYXJrIiwiYXVkIjoibWVhZG93bGFyayIsInJvbGVzIjpbInZlbmRvciJdLCJzdWIiOiJzdXBlci1ncmVhdC1TSVMiLCJqdGkiOiI4MjVkNDRiOC1hODExLTRjZmYtYmVmOC1jYzhhNjYxNjliZTAiLCJpYXQiOjE2NTg1MjQyMzUsImV4cCI6MjY1ODUyNzgzNX0.DR9-ixkAs-EUIsoRjYYGY-NLal2o8wrDYYLcKFfb3_I
+@auth_header_1=Authorization: bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJlZC1maS1tZWFkb3dsYXJrIiwiYXVkIjoibWVhZG93bGFyayIsInN1YiI6InN1cGVyLWdyZWF0LVNJUyIsImp0aSI6IjNkNTliNzVmLWE3NjItNGJhYS05MTE2LTE5YzgyZmRmOGRlMyIsImlhdCI6MTYzNjU2MjA2MCwiZXhwIjozODQ1NTQ4ODgxLCJyb2xlcyI6WyJ2ZW5kb3IiXX0.lUkNIloI8s63UkMK_hlf1iFn0P65hualLPPZcawMQZ4
+@auth_header_2=Authorization: bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJlZC1maS1tZWFkb3dsYXJrIiwiYXVkIjoibWVhZG93bGFyayIsInN1YiI6InNtYWxsLXRvd24tc2lzIiwianRpIjoiMGQwZWZmYTctYWExZi00N2E3LTk4NzktNDdhOGJkOWMzMWJiIiwiaWF0IjoxNjM2NTYyMDc5LCJleHAiOjM4NDU1NDg4ODEsInJvbGVzIjpbInZlbmRvciJdfQ.1nTCwqHcUmuUwNBF8Z2NTHqTg8HPN0mWUQ10TqOCEi0
```
stephenfuqua, line 6: Should the third header lower in the file also be updated, so that it uses the new set of credentials?

stephenfuqua (approved):

PR #64, opened 2022-07-24
Proof of concept PostgreSQL as query handler

bradbanister:
Indexing optimization will be interesting. GIN indexes are necessary on JSONB columns for search. GIN indexes
do not provide sorting, yet we need pagination support. We also need to filter by resource-name/project-name/version,
not just the JSON.
 
Problem #1: We need to query both on resource-name/project-name/version and the query parameters. Query parameters
can be handled by the GIN index on edfi_doc column, resource-name/project-name/version would be handled by the
standard btree indexes on resource-name/project-name/version columns. However, we want this to be a single index,
but you can't GIN on a string column like resource_name or BTREE on a JSONB column like edfi_doc.

Solution #1: Use the trusted btree_gin PostgreSQL extension, which extends GIN to simple column types, and create
a multi-column index on resource_name, project_name, resource_version and edfi_doc. Consider collapsing down the
resource-name/project-name/version tuple to a single concatenated column -- though this is an optimization only
relevant to "standalone PostgreSQL".

See: https://pganalyze.com/blog/gin-index

Problem #2: We need to provide pagination support. GIN indexes (including btree_gin's implementation) do not
support sorting.

Solution #2: If Meadowlark provides "next cursor"-style support instead of limit/offset, we can implement pagination on a
sortable column like documentid. This might take the form of a subquery that uses the GIN index first (ensuring the
GIN index gets used, as the query optimizer often gets the best option wrong here), then the main query sorts and limits
on documentIds, optionally starting with the next cursor token.

See: https://ivopereira.net/efficient-pagination-dont-use-offset-limit
      https://slack.engineering/evolving-api-pagination-at-slack-1c1f644f8e12
      https://dba.stackexchange.com/questions/113132/gin-index-not-used-when-adding-order-clause
      https://dba.stackexchange.com/questions/208065/why-is-limit-killing-performance-of-this-postgres-query

### DIFF
--- a/Meadowlark-js/.vscode/launch.json
+++ b/Meadowlark-js/.vscode/launch.json
@@ -33,6 +33,7 @@
       "port": 9229,
       "console": "integratedTerminal",
       "cwd": "${workspaceFolder}/services/meadowlark-fastify",
+      "envFile": "${workspaceFolder}/services/meadowlark-fastify/.env",
       "justMyCode": false
     },
     {
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/script/README.md
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/script/README.md
@@ -1,0 +1,7 @@
+These are SQL scripts for PostgreSQL backend initialization.
+
+document-index.sql - Adds a GIN "pathops" index to the edfi_doc JSONB column, improving the
+                     performance of ad-hoc document queries at the cost of slower writes.
+                     This is only necessary in a PostgreSQL "standalone" configuration,
+                     where the PostgreSQL backend is providing API search query support,
+                     as opposed to using a separate search engine.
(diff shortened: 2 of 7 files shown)


stephenfuqua:
Great line of inquiry, thank you for the investigation. I was wondering about this with our first pilot installation - simplifying so that OpenSearch or alternative is not required. The limit/offset is hard: it is a necessary part of the API. I've seen that cursor pattern elsewhere, for example with GitHub API. 

It will be worth exploring this further and thinking about the actual usage patterns in the field. For example, how many API clients really need to page through all records? Is that primarily for enrollment data, which can be provided (in the future) through a separate API service & data store (materialized views)? Is it for API extraction and sync-up? If so, can those programs adjust to use a different pattern? Or simply pull data from Kafka?

Whatever the answers, this research will be important for understanding the limitations and possibilities without making undue assumptions.

bradbanister:
Closing as this won't be merged as is, but remain as a springboard for future work on standalone configurations.

PR #65, opened 2022-07-26
[RND-276] Full Access to GET descriptors

stevenarnold-ed-fi:


Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/OwnershipSecurity.ts
```diff
@@ -22,8 +22,8 @@ export async function rejectByOwnershipSecurity(
   const functionName = 'OwnershipSecurity.rejectByOwnershipSecurity';
   Logger.info(functionName, frontendRequest.traceId, frontendRequest);
 
+  // RND-234 If it's a GET request and a descriptor, ignore ownership
   if (
-    // RND-234 If it's a GET request and a descriptor, ignore ownership
```
stephenfuqua: Can you remove `RND-234` from this? It really doesn't have anything to do with that. Not sure why someone wrote that.

stephenfuqua (commented):
Missing one requirement:
> "Get All works for all authenticated API clients"

Thus you'll need to update the opensearch backend as well. in `local.33b.http` you can see that this isn't working with the test on line 20. Or at least I assume this is the reason that line 20 is not working. 

Also makes me realize that we didn't deal with OpenSearch for full access. If easy to implement, can you fix that in this ticket? Else we can create a new ticket just for that.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/OwnershipSecurity.ts
stephenfuqua: By the way, I was accidentally looking at only the most recent commit when I wrote that. So I thought you had just moved an existing comment from before you started. Oops.

Anyway, RND-234 was the wrong one and doesn't really need to be mentioned.

Meadowlark-js/tests/http/local.ownership-tests.http
```diff
@@ -155,3 +155,32 @@ DELETE http://localhost:3000/local/v3.3b/ed-fi/educationContents/YKe0uvuOUdWgG2M
 ### Delete 2nd resource using the same authorization header that created the record - expect success.
 DELETE http://localhost:3000/local/v3.3b/ed-fi/educationContents/1vvBcOeM1WsZkfwgE-edz-dPgkHJY7uyBrEkuQ
 {{auth_header_2}}
+
```
stephenfuqua, line 158: The test on line 73 is giving me a rather unusual response with PostgreSQL as the backend, but not MongoDB:

```
HTTP/1.1 400 Bad Request
content-length: 28
Date: Wed, 27 Jul 2022 22:08:09 GMT
Connection: close

{
  "message": "Malformed body"
}
```

I doubt that it is due to any changes in this PR. I'm going to move past that and create a separate bug ticket.

stephenfuqua (approved):

PR #67, opened 2022-07-27
[RND-234] Reference extraction document path collecting must recurse

bradbanister:


Meadowlark-js/packages/meadowlark-core/src/validation/DocumentReferenceExtractor.ts
```diff
@@ -76,6 +76,29 @@ const multiZip = (...theArrays) => {
...
+/**
+ * Recursively collects JSON document paths from sub-ReferenceComponents, which represent identities that are
+ * references to other entities. In a MetaEd model, there is no limit to the number of entities that reference others
+ * as a part of their identity.
+ */
+function documentPathsFromSubReferenceComponent(
+  subReferenceComponent: ReferenceComponent,
+  referenceComponentTopLevelName: string,
+  result: string[],
+) {
```
stephenfuqua, line 88: Trivial change requested: I like functions better when they return any modified references, so that it is more clear to the consumer that the reference has in fact been modified. For that reason in C# I usually reject the use of `ref` or `out` parameters. What do you think about changing this to return the `result` object?

I think Uncle Bob has a term for this in Clean Code... avoid side-effects, I think.

stephenfuqua (commented):
The code works as desired.

Few small requests:

1. Comment about avoiding side effects.
2. Can you update `local.33b.http` to update the comment about not being able to create a section? The next request immediately after that is the same thing, but with reference validation disabled. Don't need that second version anymore, now that the first version works.
3. `DocumentIdentity.ts` also has mentions RND-234. Can you check on that comment and see if it is still needed?

stephenfuqua (commented):
Code works as desired. Small requests:

1. "avoid side effects" comment.
2. in `local.33b.http`, remove comment about RND-234 and remove the next post after that one (the one to remove is a copy with reference validation disabled)
3. Look in `DocumentIdentity.ts` to see if the comment about RND-234 can be removed. If it is still relevant, then can we solve it now, or should that be in a different ticket?

stephenfuqua (approved):

PR #72, opened 2022-07-29
RND-296 - Descriptors could be deleted when another document was referencing descriptor

stevenarnold-ed-fi:
We were not adding the descriptor references to the outRefs of the document in both PostgreSQL and MongoDB
Also added system test covering this issue


stephenfuqua (approved):
This changes works. I found a separate, pre-existing bug while testing it. Going to merge this anyway, since the bug is on the main branch.

PR #73, opened 2022-08-01
Missed update fixes

stevenarnold-ed-fi:
The original PR for RND-296 included only fixes for upsert, this adds the fixes for update as well


stephenfuqua (approved):
This was a good find, glad you caught it.

PR #78, opened 2022-08-05
[RND-290] Simplify document identity JSON representation

bradbanister:


stephenfuqua (commented):
Code looks good. "Create a courseOffering" in local.33b.http is failing with a reference validation error. Probably caused by a different PR. Will investigate before merging this ticket.

stephenfuqua (approved):

PR #83, opened 2022-08-16
client_id for claim ownership

stevenarnold-ed-fi:
Added client_id to token
updated Meadowlark to look for clientId rather than clientName
updated OAuth to return client_id in token
updated tokens for system and http tests to have client_id


Meadowlark-js/backends/meadowlark-mongodb-backend/test/integration/OwnershipSecurity.test.ts
```diff
@@ -87,7 +87,7 @@ describe('given the getById of a document owned by the requestor', () => {
   let result;
 
   const authorizationStrategy = 'OWNERSHIP_BASED';
-  const clientName = 'ThisClient';
+  const clientId = 'ThisClient';
```
stevenarnold-ed-fi, line 90: For integration tests, I kept the old client name for the clientId, for system and http tests, I added client_id to the token and base64 encoded the clientName to get a clientId

Meadowlark-js/packages/meadowlark-core/src/security/JwtAction.ts
```diff
@@ -21,13 +21,14 @@ function signingKey(): Buffer {
...
 const cachedSigningKey = memoize(signingKey);
 
-const claims = { iss: 'ed-fi-meadowlark', aud: 'meadowlark', roles: [] as string[] };
+const claims = { iss: 'ed-fi-meadowlark', aud: 'meadowlark', roles: [] as string[], client_id: '' };
 
 /*
  * Creates a standard Meadowlark Jwt.
  */
-export function createToken(vendor: string, role: string): Jwt {
+export function createToken(clientId: string, vendor: string, role: string): Jwt {
```
stevenarnold-ed-fi, line 29: For hardcoded credentials, used the client key as the client id

bradbanister (changes requested):
On several of the local.token.http tests I get 500 errors with `Cannot read properties of undefined (reading 'client_id')`

bradbanister (approved):
Looks good

PR #84, opened 2022-08-16
[RND-341] fix missing async/awaits

bradbanister:


Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Update.ts
```diff
@@ -87,11 +87,12 @@ export async function updateDocumentById(
     outRefs.push(...descriptorOutRefs);
 
     // Perform insert of references to the references table
-    outRefs.forEach(async (ref: string) => {
+    // eslint-disable-next-line no-restricted-syntax
+    for (const ref of outRefs) {
```
stevenarnold-ed-fi, line 91: Why the switch from foreach?

bradbanister: A good question. Basically, the problem is that forEach() offers no way to await what's going on inside, even if the function inside is async. It predates async/await, which is the reason for the disconnect.

Google search will turn up plenty of info, but here's a good writeup: https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971

One wrinkle is that the airbnb eslint ruleset doesn't like using for..of in general for some reason, which is the reason for the eslint-disable.

stevenarnold-ed-fi (approved):

PR #85, opened 2022-08-19
[RND-329] Change document id to separate hashes of resource info and document identity

bradbanister:


Meadowlark-js/packages/meadowlark-core/src/model/DocumentIdentity.ts
```diff
@@ -32,27 +31,38 @@ function toBase64Url(base64: string): string {
...
+}
+
+/**
+ * Returns a 224-bit document id for the given document identity, as a concatenation of two Base64Url hashes.
+ *
+ * The first 96 bits (12 bytes) are a SHAKE256 hash of the resource info.
+ * The remaining 128 bits (16 bytes) are a SHAKE256 hash of the document identity.
+ *
+ * The resulting Base64Url string is 38 characters long. The first 16 characters are the resource info hash and
+ * the remaining 22 characters are the identity hash.
```
bradbanister, line 47: Idea here is that project and resource name were previously bundled with document identity to create the hash that was the id. However, to validate against just project/resource name, I pulled them out into their own hash, and id is now the concatenation of the two.

Now meadowlark-core can validate an id is for the proper resource by comparing the project/resource with the first hash. However, there are elements of id lengths to get agreement on:

1) Total id length, currently 224 bits. This is the same as the previous length. Note that 224 was originally chosen because it was the shortest standard length for SHA-3 hashes.
2) Document identity length, currently 128 bits.  This seemed a nice round number while providing some space for the other hash. (By comparison, git commits are 160 bits and S3 etags are 128 bits.)
3) Project/resource name length, currently 96 bits by subtraction.

stephenfuqua: Does this actually improve the situation, since there are now two different hashes, and now we have risk_total = risk_1 * risk_2 < (risk_1, risk_2).

I'm not bothered by these key lengths right now.

bradbanister: I think it is "worse" (relatively speaking) just because we are allocating the bits to each piece of the id, and it's hard to know what the optimal split would be. For example, we might provide too many bits to the resource name and not enough to the document identity as compared to putting them all together.

In general I'm fine with it for now as well.

stephenfuqua: I think I understand now. Being able to set the length of the hash was really throwing me off, as I haven't seen that elsewhere. 

If I remember correctly, this `document_id` makes it feasible to look for a single key in the data store, instead of a composite key. And we base64 encode in order to save a string. We have 38 characters right now. Maybe we're being overly conservative about storage.

What if we use plain text for the "left side" (`Ed-Fi#3.3.1-b#StudentEducationAssociation`) and use the original 228 bit hash on the "right-side" of the key? (`#MhImkOIZ8zdxsrQ_Qn0pPqCwh0mY6hDFnGXy7A`).

bradbanister: Your suggestion is very similar to my attempt at an "easy" solution. However, it makes for oddly repetitive client requests if is reflected all the way up to the client (not suggesting this) e.g. http://localhost:3000/local/v3.3b/ed-fi/StudentEducationAssociation/Ed-Fi#3.3.1-b#StudentEducationAssociation#MhImkOIZ8zdxsrQ_Qn0pPqCwh0mY6hDFnGXy7A, so it think pushing that up to the client is out.

But if that's out, then this validation must be pushed down into every backend, which is where I was seeing the code get overly complicated and goofy (a technical term) on my 2nd attempt. The code is a quite simple meadowlark-core middleware in this latest implementation.

Meadowlark-js/.yarnrc.yml
```diff
@@ -10,12 +10,6 @@ npmScopes:
   edfi:
     npmRegistryServer: "https://www.myget.org/F/ed-fi/npm/"
 
-packageExtensions:
-  inquirer-autocomplete-prompt@*:
-    peerDependenciesMeta:
-      inquirer:
-        optional: true
-
```
bradbanister, line 18: Outdated yarn directive

Meadowlark-js/packages/meadowlark-core/src/handler/FrontendFacade.ts
```diff
@@ -36,16 +37,45 @@ function methodWithBodyStack(): MiddlewareStack {
...
-// Middleware stack builder for deletes - no body
+// Middleware stack builder put, body and id
+function putStack(): MiddlewareStack {
+  return R.once(
+    R.pipe(
+      authorize,
+      R.andThen(parsePath),
+      R.andThen(parseBody),
+      R.andThen(resourceValidation),
+      R.andThen(resourceIdValidation),
```
bradbanister, line 48: Inserting id validation into the stacks

Meadowlark-js/packages/meadowlark-core/src/middleware/ValidateResourceIdMiddleware.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function resourceIdValidation({
+  frontendRequest,
+  frontendResponse,
+}: MiddlewareModel): Promise<MiddlewareModel> {
+  // if there is a response already posted, we are done
+  if (frontendResponse != null) return { frontendRequest, frontendResponse };
+  writeRequestToLog(moduleName, frontendRequest, 'resourceIdValidation');
+
+  // If no resource id in request, there is nothing to validate
+  if (frontendRequest.middleware.pathComponents.resourceId == null) return { frontendRequest, frontendResponse: null };
```
bradbanister, line 24: Just in case it goes in a stack by accident without resource ids

Meadowlark-js/packages/meadowlark-core/src/middleware/ValidateResourceIdMiddleware.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  if (
+    !isDocumentIdValidForResource(
+      frontendRequest.middleware.pathComponents.resourceId,
+      frontendRequest.middleware.resourceInfo,
+    )
+  ) {
```
bradbanister, line 31: Middleware is this simple

Meadowlark-js/packages/meadowlark-core/src/model/DocumentIdentity.ts
```diff
@@ -32,28 +31,41 @@ function toBase64Url(base64: string): string {
...
+ * Hashes a string with the SHAKE256, returning a Base64 hash with the specified length given in bytes
  */
-export function documentIdForDocumentIdentity(
-  { projectName, resourceName, isDescriptor }: BaseResourceInfo,
-  documentIdentity: DocumentIdentity,
-): string {
+function toHash(data: string, lengthInBytes: number): string {
+  return crypto.createHash('shake256', { outputLength: lengthInBytes }).update(data).digest('base64');
+}
+
```
bradbanister, line 39: SHAKE hashing lets you specify output length

Meadowlark-js/packages/meadowlark-core/src/validation/DocumentIdValidator.ts
```diff
@@ -0,0 +1,22 @@
...
+export function isDocumentIdWellFormed(documentId: string): boolean {
+  return /^[^\s/+]{38}$/g.test(documentId);
+}
+
+/**
+ * Returns true if resource info hash matches resource info portion of document id
+ */
+export function isDocumentIdValidForResource(documentId: string, resourceInfo: BaseResourceInfo): boolean {
+  return documentId.startsWith(resourceInfoHashFrom(resourceInfo));
+}
```
bradbanister, line 22: Id validation becomes a one-liner

Meadowlark-js/packages/meadowlark-core/src/validation/DocumentIdValidator.ts
```diff
@@ -0,0 +1,22 @@
...
+import { resourceInfoHashFrom } from '../model/DocumentIdentity';
+import type { BaseResourceInfo } from '../model/ResourceInfo';
+
+/**
+ * Document Ids are 38 character Base64Url strings. No whitespace, plus or slash allowed
+ * Example valid id: 02pe_9hl1wM_jO1vdx8w7iqmhPdEsFofglvS4g
+ */
+export function isDocumentIdWellFormed(documentId: string): boolean {
+  return /^[^\s/+]{38}$/g.test(documentId);
+}
```
bradbanister, line 15: Moved the id validation to its own module

stephenfuqua (approved):

PR #87, opened 2022-08-22
[RND-309] PostgreSQL row locking

stevenarnold-ed-fi:


Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
```diff
@@ -8,11 +8,16 @@ import format from 'pg-format';
 
 /**
  * Function that returns a SQL query for checking whether references exist for a given list of document ids
+ * This has a secondary function of locking this row for proper updating and deleting
  * @param documentId the document to check for existence
  * @returns SQL query to check for existence ids
  */
 export function existenceIdsForDocument(documentId: string): string {
```
stephenfuqua, line 15: I wonder about renaming this so that it is more self-explanatory to the reader. An idea:

```suggestion
export function getReferenceIdentifiers(documentId: string): string {
```

Is that truly better? Not sure. I still don't like the table name `existence`. Now I'm wondering about changing to `references`. But that would be for another ticket, not this one. 

Let's not change the name right now. Leaving this comment in place though, to spur discussion before creating a ticket.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
```diff
@@ -8,11 +8,16 @@ import format from 'pg-format';
...
-  return format(`SELECT existence_id FROM meadowlark.existence WHERE document_id = %L`, documentId);
+  return format(`SELECT existence_id FROM meadowlark.existence WHERE document_id = %L FOR SHARE NOWAIT`, documentId); //
+}
+
+export function existenceCheckForDocument(documentId: string): string {
```
stephenfuqua, line 19: Can we make this more action-oriented? Maybe something like

```suggestion
export function lookForExistingDocument(documentId: string): string {
```

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
```diff
@@ -46,7 +51,10 @@ export function documentOwnershipByIdSql(documentId: string): string {
  * @returns SQL query string to retrieve references
  */
 export function checkForReferencesByDocumentId(documentId: string[]): string {
```
stephenfuqua, line 53: This function name: another reason to consider renaming `existence` to `references`.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
stevenarnold-ed-fi: I'm not really happy with the name of the existence table either. Something to keep in mind though, we do already have a references table in the Postgres implementation.

stephenfuqua: Ah, I'm getting them confused - because I can never remember what "existence" means! All the more reason to change it. Now I remember that it is for the superclass identification, not for "foreign keys". So changing this function name to use the word "references" would be entirely inappropriate.

stephenfuqua: I created a ticket for it.

stephenfuqua:
There's a really strange problem here:

1. Drop all tables - start with a clean slate.
2. Start meadowlark
3. Open `local.33b.superclass.http`
4. Start running from top to bottom.
5. Line 129 fails with 409 and a message indicating that there is a SchoolCategory that references this School. But that's backwards, the _School_ references the _SchoolCategoryDescriptor_. So there is no reason that deleting the school should fail.

PR #95, opened 2022-08-31
[RND-349] Support renamed schoolYear

stephenfuqua:
I'm planning to add some unit tests around this before moving to QA.


Meadowlark-js/packages/meadowlark-core/src/validation/DocumentIdentityExtractor.ts
```diff
@@ -32,8 +32,18 @@ type NullableTopLevelEntity = { superclass: TopLevelEntity | null };
  */
 function singleIdentityFrom(property: EntityProperty, body: object, documentPath: string[]): DocumentIdentity {
   const { apiMapping } = property.data.meadowlark as EntityPropertyMeadowlarkData;
-  const documentPathAsString: string = [...documentPath, apiMapping.fullName].join('.');
-  const elementValue: string | undefined = R.path([...documentPath, apiMapping.fullName], body);
+
+  let path: string[];
+
+  if (property.type === 'schoolYearEnumeration' && property.roleName) {
+    path = [...documentPath, 'schoolYear'];
```
stephenfuqua, line 39: I could have used `property.metaEdName` or something like that here, but chose to hard code `schoolYear` so that it is more explicit & obvious.

Without this clause, the path was coming out as `graduationSchoolYearTypeReference.graduationSchoolYear` because the `apiMapping.fullName == 'graduationSchoolYear'`.

Meadowlark-js/packages/meadowlark-core/src/validation/DocumentIdentityExtractor.ts
```diff
@@ -62,7 +72,12 @@ function documentIdentitiesFrom(
   if (isReferenceElement(identityReferenceComponent)) {
     // SchoolYearEnumerations are an API one-off expressed as two levels in the document body
     if (identityReferenceComponent.sourceProperty.type === 'schoolYearEnumeration') {
-      documentPath.push('schoolYearTypeReference');
+      if (identityReferenceComponent.sourceProperty.roleName) {
```
stephenfuqua, line 75: I need to jog my memory on the better way to handle checking if there is a value here. Don't tell me yet! I need to look into and get it to sink into my memory.

Meadowlark-js/packages/meadowlark-core/src/validation/DocumentIdentityExtractor.ts
```diff
@@ -32,8 +32,18 @@ type NullableTopLevelEntity = { superclass: TopLevelEntity | null };
  */
 function singleIdentityFrom(property: EntityProperty, body: object, documentPath: string[]): DocumentIdentity {
   const { apiMapping } = property.data.meadowlark as EntityPropertyMeadowlarkData;
-  const documentPathAsString: string = [...documentPath, apiMapping.fullName].join('.');
-  const elementValue: string | undefined = R.path([...documentPath, apiMapping.fullName], body);
+
+  let path: string[];
+
+  if (property.type === 'schoolYearEnumeration' && property.roleName) {
```
stephenfuqua, line 38: Looking at this again, I should have used the same string "null or empty" check that I employed on line 76 below.

bradbanister, line 38: Not wrong to do a roleName check this way but the type of EntityProperty.roleName (like almost everything in the MetaEd model) is not nullable. Since the string is not nullable, the absence of a string value in the MetaEd model is the empty string, which simplifies these kind of checks.

Meadowlark-js/packages/meadowlark-core/src/validation/DocumentIdentityExtractor.ts
```diff
@@ -62,7 +72,12 @@ function documentIdentitiesFrom(
   if (isReferenceElement(identityReferenceComponent)) {
     // SchoolYearEnumerations are an API one-off expressed as two levels in the document body
     if (identityReferenceComponent.sourceProperty.type === 'schoolYearEnumeration') {
-      documentPath.push('schoolYearTypeReference');
+      const { roleName } = identityReferenceComponent.sourceProperty;
+      if (roleName != null && roleName.trim().length > 0) {
```
bradbanister, line 76: As above, not wrong to do a roleName check this way but the type of EntityProperty.roleName (like almost everything in the MetaEd model) is not nullable. Since the string is not nullable, the absence of a string value in the MetaEd model is the empty string, which simplifies these kind of checks.

bradbanister (commented):
Looks good. I'll leave to @stevenarnold-ed-fi to approve. Only comment is inline about how MetaEd does not believe in nulls 😉 in almost all model values, so they are non-nullable types. For strings the default value is an empty string. Non-nullable types are a simple feature of TypeScript which has been fantastic for MetaEd.

Historically, the reason for this stems from crash avoidance while linting on the fly in the IDE. While typing, the .metaed file is usually not well-formed from a parser perspective so parts of the MetaEd model after parse are missing. By requiring all model elements to have a not-null default value, we avoided bogging down the code with null checks that would otherwise be required everywhere in all plugins just to handle invalid .metaed files. Miss even one null check and the IDE will report a crash on every keystroke until the file is well-formed again.

In reality, we don't care about the parts of the .metaed files that aren't well formed. The parser reports that. We just want to be able to provide linting feedback on the parts that _are_ well-formed.

Meadowlark-js/packages/meadowlark-core/src/validation/DocumentIdentityExtractor.ts
stephenfuqua: done

PR #96, opened 2022-08-31
RND-345 - Remove hard coded tokens from http test files

stevenarnold-ed-fi:
Also added an existence check to update, so that an error response is returned when you try to insert on an update

### DIFF
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Update.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Update.ts
@@ -13,10 +13,11 @@ import {
 } from '@edfi/meadowlark-core';
 import type { PoolClient, QueryResult } from 'pg';
 import {
-  insertAliasSql,
-  deleteAliasesForDocumentSql,
-  deleteOutboundReferencesOfDocumentSql,
   documentInsertOrUpdateSql,
+  findAliasIdsForDocumentSql,
+  deleteAliasesForDocumentSql,
+  insertAliasSql,
+  deleteOutboundReferencesOfDocumentSql,
   insertOutboundReferencesSql,
 } from './SqlHelper';
 import { validateReferences } from './ReferenceValidation';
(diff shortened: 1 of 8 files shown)


Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Update.ts
```diff
@@ -34,6 +35,13 @@ export async function updateDocumentById(
   try {
     await client.query('BEGIN');
 
+    const recordExistsResult = await client.query(findAliasIdsForDocumentSql(id));
+
+    if (recordExistsResult.rowCount === 0) {
```
stevenarnold-ed-fi, line 40: Bug that Stephen found, without checking for the existence of a record on an update, reference and alias table insertions but no document table insert

Meadowlark-js/tests/http/local.deleteissue.http
```diff
@@ -115,7 +115,16 @@ reference-validation: true
         }
     ]
 }
-###
+### This delete will fail because the school is referenced by the location and the ed-org is referenced 
+# by the course
 @schoolDocumentlocation = {{school124.response.headers.location}}
 DELETE http://localhost:3000{{schoolDocumentlocation}}
+{{auth_header}}
+
```
stevenarnold-ed-fi, line 123: these tests got us to the point of showing that once you references a school, you could not delete it, but left it there. Added a delete for location and course objects allowing the school to be deleted

stevenarnold-ed-fi:
I left the hardcoded tokens in the token http file that were there to show the response from bad tokens, since we can't replicate that with a call to the oauth endpoint

Meadowlark-js/tests/http/local.ownership-tests.http
```diff
@@ -1,11 +1,42 @@
 ### Set `ACCESS_TOKEN_REQUIRED=true` environment variable before `yarn start:local`
 # Be sure to load all descriptors into the database first.
 
+### Authenticate
```
stephenfuqua, line 4: @stevenarnold-ed-fi in this file, we have "expect 403" in several places, inconsistent with `local.token.http` as you've noted. But it makes me wonder if we differentiated between 404 for unauthenticated users and 403 for authenticated users.

stephenfuqua (approved):

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Update.ts
stephenfuqua: Give yourself credit here! I saw it, but didn't know what was going on and moved past. You saw it and dug into the details.

PR #97, opened 2022-09-02
[RND-330] Disallow subclasses with same superclass identity

bradbanister:


Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -20,6 +20,31 @@ export async function upsertDocument(
...
+
+        if (superclassAliasIdInUse) {
+          Logger.debug(
+            `mongodb.repository.Upsert.upsertDocument: Upserting document id ${id} failed due another subclass with the same identity`,
+            traceId,
+          );
+
+          upsertResult = {
+            response: 'INSERT_FAILURE_CONFLICT',
+            failureMessage: `Insert failed: the identity is in use by another resource with the same superclass`,
```
stephenfuqua, line 40: Here's how the ODS/API responds when you try to create an LEA with an existing SchoolId:

```
{
  "message": "A natural key conflict occurred when attempting to create a new resource 'EducationOrganization' with a duplicate key.  The duplicated columns and values are [EducationOrganizationId] (5) This is likely caused by multiple resources with the same key in the same file. Exactly one of these resources was inserted."
}
```

How difficult would it be to detect what the superclass is (e.g. EducationOrganization) and include that in the message, instead of the word "superclass"? I don't think we need to perfectly match up the error message.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
```diff
@@ -33,6 +33,20 @@ export function findAliasIdsForDocumentSql(documentId: string): string {
   return format(`SELECT alias_id FROM meadowlark.aliases WHERE document_id = %L FOR SHARE NOWAIT`, documentId);
 }
 
+/**
+ * Returns the SQL query to find the existence of an alias id. Alias ids include the document id
+ * itself along with any superclass variations that the document satisifies. For example, a School is a subclass
+ * of EducationOrganization, so each School document has an additional alias id as an EducationOrganization.
+ *
+ * This function does NOT read-for-write lock.
```
stephenfuqua, line 41: I agree with that. There is a tiny possibility of two competing clients racing to post a record first - unless we lock the entire table, we can't prevent that. So be it.

stephenfuqua (approved):

PR #98, opened 2022-09-02
[RND-326] Adding total count from open search

stevenarnold-ed-fi:


Meadowlark-js/packages/meadowlark-core/src/handler/Query.ts
```diff
@@ -14,6 +14,7 @@ import { FrontendRequest } from './FrontendRequest';
 import { FrontendResponse } from './FrontendResponse';
 
 const moduleName = 'Query';
+const TOTAL_COUNT_HEADER_NAME: string = 'total-count';
```
stevenarnold-ed-fi, line 17: Added - so that it matched most existing headers. totalCount would become 'totalcount' in the headers, which looks weird to me.

Meadowlark-js/packages/meadowlark-core/src/handler/Query.ts
```diff
@@ -54,9 +55,15 @@ export async function query(frontendRequest: FrontendRequest): Promise<FrontendR
 
   writeDebugStatusToLog(moduleName, frontendRequest, 'query', 200);
   const body = JSON.stringify(documents);
+
+  const headers = {
+    ...frontendRequest.middleware.headerMetadata,
+    [TOTAL_COUNT_HEADER_NAME]: result.totalCount?.toString() ?? '0',
+  };
```
stephenfuqua, line 62: There's one "problem" with this: the ODS/API only reports the total-count header if you use the query string `totalCount=true`. We're reporting in all cases. That said, it doesn't really matter. We didn't make any extra effort to get this information from OpenSearch, so let's just include it. So I'm fine with this.

stephenfuqua (changes requested):
Code looks good. Would like to see unit test coverage too: both `Query` and `QueryOpensearch` have tests files already, so should be easy to extend the coverage.

stephenfuqua (approved):

PR #99, opened 2022-09-07
[RND-321] - Move validation from header to token

stevenarnold-ed-fi:


stevenarnold-ed-fi:
Rather than having both authorization strategy and validation in the same role, it felt more correct to add a second role that was auth related. so client 4 has roles of ['vendor', 'assessment']. If we don't want to go this route and would rather have them all in one, it's only about 15 minutes of work to switch.

Meadowlark-js/packages/meadowlark-core/src/security/Security.ts
```diff
@@ -13,12 +13,15 @@ export type Security = {
...
   clientId: string;
+  // Whether to validate resources or not, currently pulled from role type 'assessment' from JWT
+  validateResources: boolean;
 };
 
 export function newSecurity(): Security {
   return {
     authorizationStrategy: 'UNDEFINED',
     clientId: 'UNKNOWN',
+    validateResources: true,
```
stevenarnold-ed-fi, line 24: Since we don't have validation in the header, and it needs to be in the FrontendRequest, this felt like the most logical place for validation

Meadowlark-js/packages/meadowlark-core/src/middleware/AuthorizationMiddleware.ts
```diff
@@ -31,6 +31,7 @@ export async function authorize({ frontendRequest, frontendResponse }: Middlewar
   frontendRequest.middleware.security = {
     authorizationStrategy: jwtStatus.authorizationStrategy as AuthorizationStrategy,
     clientId: jwtStatus.clientId ?? 'UNKNOWN',
+    validateResources: !jwtStatus.roles.includes('assessment'),
```
stephenfuqua, line 34: Since this isn't strictly a "security" thing, what do you think about moving the `validateResources` property to the `middleware` object?

Meadowlark-js/tests/http/local.33b.http
```diff
@@ -16,6 +16,22 @@ content-type: application/json
 
 ###
 @authToken1 = {{client1.response.body.$.access_token}}
+###
+
+### Authenticate client4 (this client has validation turned off, this is a replacement for reference-validation:false)
```
stephenfuqua, line 21: After this PR, the "this is ..." comment won't make sense to readers.
```suggestion
### Authenticate client4 - assessment vendor for bypassing reference validation
```

Meadowlark-js/tests/http/local.33b.superclass.http
```diff
@@ -18,6 +18,20 @@ content-type: application/json
 ###
 @auth_header=Authorization: bearer {{client1.response.body.$.access_token}}
 
+### Authenticate client4 (this client has validation turned off, this is a replacement for reference-validation:false)
```
stephenfuqua, line 21: same comment as above

Meadowlark-js/packages/meadowlark-core/src/middleware/AuthorizationMiddleware.ts
stevenarnold-ed-fi: Yep, can do.

Meadowlark-js/tests/http/local.33b.http
stevenarnold-ed-fi: can do

PR #105, opened 2022-09-22
Setting jest threshold value to a minimum where still passes validation

andonyns:
Setting values for jest coverage to a value that still passes but is close to the current coverage.

If the new changes start to decrease the coverage, the build will fail until adding new tests


bradbanister (approved):
Looks good!

PR #108, opened 2022-09-23
[RND-360] Creating workflow for report generation

andonyns:
This allows PR's from forks to be able to run the testing without getting errors generating reports.
* Moving report generation to a different workflow to avoid failures from community contributions
* Setting up working directory for test workflow
* Adding reports workflow to create test reports of execution
* Using matrix to generate reports

### DIFF
--- a/.github/workflows/reports.yml
+++ b/.github/workflows/reports.yml
@@ -3,30 +3,34 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
-name: 'Reports'
+name: Reports
 on:
   workflow_run:
-    workflows: [ 'Test' ]
+    workflows: ["Test"]
     types:
       - completed
 
 jobs:
-    report:
-        runs-on: ubuntu-latest
-        strategy:
-          fail-fast: false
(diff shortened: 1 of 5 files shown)


.github/workflows/reports.yml
```diff
@@ -0,0 +1,32 @@
...
+name: 'Reports'
+on:
+  workflow_run:
+    workflows: [ 'Test' ]
+    types:
+      - completed
+
+jobs:
+    report:
+        runs-on: ubuntu-latest
```
stephenfuqua, line 15: It is a small thing... preference for 2 spaces as the indent instead of 4 in YML files. We have that specified in the `.editorconfig` file at the root of the project. Do you have the editorconfig extension for VS Code? It will set VS Code to use 2 spaces automatically.

andonyns: I will review this. I think I do but not sure why it didn't do it

andonyns: I found that there are two `.editorconfig`  files, one in the root and one in Meadowlark-js. I think this is needed depending if you open VS Code from the root or from the Meadowlark-js folder. Those two files had different setup so I matched the settings.

stevenarnold-ed-fi (approved):

PR #112, opened 2022-10-03
[RND-363] Migrate postman tests from ODS / API into Meadowlark

andonyns:
Changes:
- Copying collections from [ODS repo](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/tree/main/Postman%20Test%20Suite) into Meadowlark.
- Modify all URLs to point to Meadowlark.
- Update authentication to work against Meadowlark endpoint.
- Fixing body data type validation errors. (not executed for all endpoints).
- Create meadowlark specific environment.

Most tests are failing from execution errors, that will be addressed in follow up tickets.


stephenfuqua:
Can you rename each file with `Meadowlark Ed-Fi API` instead of `Ed-Fi ODS/API`?

Emphasizing the difference in products here, and making sure that the names don't conflict when loading both into postman at the same time.

stephenfuqua:
Most of these files are not needed because Meadowlark does not support the features, and might not ever support them. Please remove:

* AuthorizationTests Multiple Key/Secret - the authorization model is very different, and changing these tests to fit the Meadowlark authorization is probably not worthwhile. The .http files cover authorization very well.
* AuthorizationTests - the namespace-based definitely are not needed. The other ones might be useful, but I don't think we really need them, and again they'll probably take a lot of time to fix up for use with Meadowlark.
* Multiple Authorization strategy - we only have one strategy
* Profile Test - profiles are not supported

Might be worth keeping:

* GetByExample Tests - although will probably need a little modification, as I think the querystring will be `?studentReference.studentUniqueId=xyz` instead of `?studentUniqueId=xyz`.
* Multiple Edorg types - I have not studied closely enough yet to understand what these are really doing. They look very interesting and we should try them out, but carefully / slowly - one scenario at a time. Probably needs to be me or Brad who looks at these because we'll have the best understanding of the relevance.
* ResponseTests are mostly good. We don't support etag right now, so those are not immediately relevant. We might in the future though.
* The one simply called "Ed-Fi ODS/API Integration Test Suite" has some good tests. Seems very similar to some of our .http files. Good to keep some of those, though we might want to rename the file to something more creative.
* Record Ownership - this is the authorization strategy we actually use. However, the implementation is different enough that I'm not sure these tests will make any sense. Still, it will be worthwhile to look at them more closely to see if there is anything valuable to us.

PR #113, opened 2022-10-04
[RND-299] - Harden PostgreSQL error handling

stevenarnold-ed-fi:
These are all checks to make sure that the objects returned from a query have a rows object before access. I don't see a problem with these in general, but... 

If a database or table gets deleted while Meadowlark running the query will error, and wind up in the catch block which will end with a 500 error to the client, it will never hit the rows objects.

If a table is deleted and Meadowlark is restarted, there is no error, any query that attempts to hit a non-existent table will return empty and thus most of the time act like a 404.

If the database gets deleted and Meadowlark is restarted, the database is generated.

### DIFF
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
@@ -29,7 +29,7 @@ export async function deleteDocumentById(
       const documentAliasIdsResult: QueryResult = await client.query(findAliasIdsForDocumentSql(id));
 
       // All documents have alias ids. If no alias ids were found, the document doesn't exist
-      if (documentAliasIdsResult?.rowCount === 0) {
+      if (documentAliasIdsResult.rowCount == null || documentAliasIdsResult.rowCount === 0) {
         await client.query('ROLLBACK');
         deleteResult = { response: 'DELETE_FAILURE_NOT_EXISTS' };
         return deleteResult;
@@ -40,7 +40,13 @@ export async function deleteDocumentById(
 
       // Find any documents that reference this document, either it's own id or an alias
       const referenceResult = await client.query(findReferencingDocumentIdsSql(documentAliasIds));
-      const references = referenceResult.rows.filter((ref) => ref.document_id !== id);
+
+      if (referenceResult.rows == null) {
+        await client.query('ROLLBACK');
+        return deleteResult;
+      }
+
+      const references = referenceResult?.rows.filter((ref) => ref.document_id !== id);
 
       // If this document is referenced, it's a validation failure
       if (references.length > 0) {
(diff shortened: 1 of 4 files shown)


Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Get.ts
```diff
@@ -11,6 +11,11 @@ export async function getDocumentById({ id }: GetRequest, client: PoolClient): P
   try {
     const queryResult: QueryResult = await client.query(findDocumentByIdSql(id));
```
stephenfuqua, line 12: I put a breakpoint on here and ran in debug mode. When it paused here, I renamed `documents` to `documents_2`. And it _did_ throw an appropriate error that I can see in the debugger. However, it just gave me a 500 response and did not log the error.

image

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/OwnershipSecurity.ts
```diff
@@ -44,6 +44,10 @@ export async function rejectByOwnershipSecurity(
   try {
     const result: QueryResult = await client.query(findOwnershipForDocumentSql(id));
```
stephenfuqua, line 45: I was able to see the same error as shown in screenshot above when I put a breakpoint here.

stephenfuqua (changes requested):

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Get.ts
stevenarnold-ed-fi: This is only true during the runtime of Meadowlark, once it's restarted, you will no longer see an error

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -43,6 +44,10 @@ export async function deleteDocumentById(
 
       if (referenceResult.rows == null) {
         await client.query('ROLLBACK');
+        const errorMessage = `deleteDocumentById: Error determining documents referenced by ${id},
+        a null result set was returned`;
```
stephenfuqua, line 48: I like lines that aren't very wide, but not at the expensive of having a newline show up in a log file 😆 
```suggestion
        const errorMessage = `deleteDocumentById: Error determining documents referenced by ${id}, a null result set was returned`;
```

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -61,6 +66,10 @@ export async function deleteDocumentById(
 
         if (referringDocuments.rows == null) {
           await client.query('ROLLBACK');
+          const errorMessage = `deleteDocumentById: Error retrieving documents referenced by ${id},
+          a null result set was returned`;
```
stephenfuqua, line 70:
```suggestion
          const errorMessage = `deleteDocumentById: Error retrieving documents referenced by ${id}, a null result set was returned`;
```

stephenfuqua (approved):

PR #116, opened 2022-10-11
[RND-379] School year validation

stephenfuqua:
* Needs more tests
* Need to remove the `schoolYearType` resource from the API surface

### DIFF
--- a/Meadowlark-js/packages/meadowlark-core/src/Environment.ts
+++ b/Meadowlark-js/packages/meadowlark-core/src/Environment.ts
@@ -6,9 +6,9 @@
 export type EnvironmentVariable = 'SIGNING_KEY' | 'TOKEN_URL' | 'ALLOWED_SCHOOL_YEARS';
 
 // Local cache to be used for storing parsed environment variables.
-const cache: { [key: string]: any } = {};
+let cache: { [key: string]: any } = {};
 
-/*
+/**
  * Gets a value from the environment variables, throwing an error if not found.
  */
 export function getValueFromEnvironment(key: EnvironmentVariable): string {
@@ -21,14 +21,14 @@ export function getValueFromEnvironment(key: EnvironmentVariable): string {
   throw new Error(`Environment variable '${key}' has not been setup properly.`);
 }
 
-/*
+/**
  * Insert or replace a value in the environment cache by key name.
  */
 export function updateCache(key: EnvironmentVariable, value: any): void {
   cache[key] = value;
 }
 
-/*
+/**
  * Retrieve a value from the environment cache by key name.
  */
 export function getFromCache<T>(key: EnvironmentVariable): T | undefined {
@@ -39,16 +39,16 @@ export function getFromCache<T>(key: EnvironmentVariable): T | undefined {
   return undefined;
 }
 
-/*
+/**
  * Remove a key from the environment cache.
  */
 export function removeFromCache(key: EnvironmentVariable): void {
   delete cache[key];
 }
 
-/*
+/**
  * Remove all keys from the environment cache.
  */
 export function clearCache(): void {
-  cache.clear();
+  cache = {};
 }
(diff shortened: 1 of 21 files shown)


Meadowlark-js/packages/meadowlark-core/src/Environment.ts
```diff
@@ -14,3 +20,35 @@ export function getValueFromEnvironment(key: EnvironmentVariable): string {
 
   throw new Error(`Environment variable '${key}' has not been setup properly.`);
 }
+
+/*
```
stephenfuqua, line 24: Because I'm parsing the school years out from a single string, I wanted the ability to cache them for quick lookup next time they're needed.

Meadowlark-js/packages/meadowlark-core/src/handler/FrontendFacade.ts
```diff
@@ -33,8 +34,9 @@ function postStack(): MiddlewareStack {
       R.andThen(parsePath),
       R.andThen(parseBody),
       R.andThen(resourceValidation),
-      R.andThen(metaeEdModelFinding),
+      R.andThen(metaEdModelFinding),
```
stephenfuqua, line 37: Random typo fix

Meadowlark-js/packages/meadowlark-core/src/validation/SchoolYearValidator.ts
```diff
@@ -0,0 +1,54 @@
...
+  const config = getValueFromEnvironment(ALLOWED_SCHOOL_YEARS);
+
+  const list = config
+    .split(',')
+    .map((y: string) => (/^\d+$/.test(y.trim()) ? Number.parseInt(y, 10) : NaN))
+    .filter((n: number) => !Number.isNaN(n));
+
+  updateCache(ALLOWED_SCHOOL_YEARS, list);
+
+  return list;
```
stephenfuqua, line 31: Might be worthwhile, at some point, to refactor this so that some of this code moves into the `Environment` module, perhaps with a signature like:

```
getConfigValue<T>(key, mappingFunction): T
```

Meadowlark-js/packages/meadowlark-core/src/validation/SchoolYearValidator.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * Validates any fields of type `SchoolYear` by comparing to the list of configured school years.
+ *
+ * Validates the request body for the resource. If invalid, returns an error message.
+ */
+export function validateDocument(body: SchoolYearReferenceDocument): string {
+  const { schoolYear } = (body as SchoolYearReferenceDocument).schoolYearTypeReference as SchoolYearEnumerationDocument;
```
stephenfuqua, line 40: Is this the best way to extract the `schoolYear` value?

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/src/enhancer/HasSchoolYearEnhancer.ts
```diff
@@ -0,0 +1,34 @@
...
+    'association',
+    'domainEntitySubclass',
+    'associationSubclass',
+    'common',
+  ).forEach((entity: ModelBase) => {
+    const meadowlarkEntity = entity.data.meadowlark as EntityMeadowlarkData;
+    meadowlarkEntity.hasSchoolYear = meadowlarkEntity.collectedProperties.some(
+      (property: CollectedProperty) => property.property.type === 'schoolYearEnumeration',
+    );
+  });
```
stephenfuqua, line 28: Is this the right list of entity types?

Note: this does not dive into reference types. That concerned me momentarily... but then I realized we really only need to validate any `schoolYearReference` directly on the object. Reference validation will handle any nested references.

For example, a bad year in `Session` keeps that Session from being saved.

A bad year in `CourseOffering.Session` will not be rejected by the Year validation - but it _will_ be rejected because the referenced ` Session` does not exist.

stephenfuqua:
Replacing this with a new PR

PR #120, opened 2022-10-17
[RND-364] Add Initial set of e2e Tests

andonyns:
- Add setup to use supertest.
- Migrate local.http tests to supertest using scenarios.
- Create test structure.
- Move postman tests to 'old' folder
- Add initial set of e2e tests.
- Add recommendation and setup for jest extension
- Current setup has 16 tests for different scenarios that take care of their own setup and teardown.

**Note:**
VSCode Test Extension shouldn't be used to run the e2e nor the system tests all at the same time, because the extension runs the tests in parallel by default, which can make those two types of tests to fail since some things should be created in a given order. This works correctly to run tests individually.


Meadowlark-js/tests/e2e/Locations.test.ts
```diff
@@ -0,0 +1,104 @@
...
+      let schoolId: number;
+      let schoolLocation: string;
+      let location: string;
+
+      beforeAll(async () => {
+        schoolId = 100;
+        schoolLocation = await createSchool(schoolId);
+      });
+
+      it('should add location with valid school', async () => {
```
stephenfuqua, line 41:
```suggestion
      it('accepts the request', async () => {
```

Meadowlark-js/tests/e2e/Locations.test.ts
```diff
@@ -0,0 +1,104 @@
...
+            },
+            maximumNumberOfSeats: 20,
+            optimalNumberOfSeats: 10,
+          },
+        });
+
+        await rootURLRequest
+          .get(location)
+          .auth(await getAccessToken(Clients.Vendor1), { type: 'bearer' })
+          .expect(200);
```
stephenfuqua, line 57: Shouldn't this be 201?

Meadowlark-js/tests/e2e/Locations.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      afterAll(async () => {
+        await deleteResourceByLocation(location);
+
+        await deleteResourceByLocation(schoolLocation);
+      });
+    });
+  });
+
+  describe('without strict validation', () => {
```
stephenfuqua, line 68:
```suggestion
  describe('given a token with relaxed validation', () => {
```

Meadowlark-js/tests/e2e/Locations.test.ts
andonyns: you're right. I will review this one

Meadowlark-js/tests/e2e/functions/Shared.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { chance, rootURLRequest } from '../Setup';
+import { Clients, getAccessToken } from './Credentials';
+import { createResource } from './Resources';
+
+export function generateRandomId(length = 12): string {
+  return chance.hash({ length });
+}
+
+/* TEMPORARY LOCATION */
```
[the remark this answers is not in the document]
andonyns: I think we can move this functions that create resources to its own file. 
I think of a descriptors class where we can add functions to CRUD Descriptors, that way, when an scenario requires to use the resource, just calls the method to reuse the function.

Meadowlark-js/tests/e2e/Locations.test.ts
andonyns: This is the request to verify that the location is correct after the resource was created, this should be a 200

Meadowlark-js/tests/e2e/EducationOrganizations.test.ts
```diff
@@ -0,0 +1,21 @@
...
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+import { getAccessToken, Clients } from './functions/Credentials';
+import { baseURLRequest } from './Setup';
+
+describe('Education Organizations', () => {
+  it("does not allow a post request", async () => {
+    await baseURLRequest
```
stephenfuqua, line 11: Nice warning above!

Meadowlark-js/tests/e2e/Locations.test.ts
stephenfuqua: OK that makes sense now. Shouldn't there be an assert for 201 response code?

andonyns: Line 44 calls the method createResource, this method verifies the 201 and returns the location

Meadowlark-js/tests/e2e/ResourcesReferenceValidation.test.ts
```diff
@@ -0,0 +1,106 @@
...
+            classroomIdentificationCode: 'string',
+            schoolReference: {
+              schoolId: 99,
+            },
+            maximumNumberOfSeats: 20,
+            optimalNumberOfSeats: 10,
+          })
+          .expect(400)
+          .then((response) => {
+            expect(response.body.message).toContain('Resource School is missing identity');
```
stephenfuqua, line 28: That's a terrible error message and we'll need to change it. Let's leave it alone for right now, but just be aware that it will change in the future. For the next round of tests, might want to simply use `expect(response.body).toHaveProperty('message')` (I think that's the syntax). No need to change right now though.

stephenfuqua (commented):
I think these are my final comments

stephenfuqua (approved):

PR #121, opened 2022-10-19
[RND-335] Token introspection

bradbanister:


Meadowlark-js/packages/meadowlark-authz-server/src/handler/RequestToken.ts
```diff
@@ -4,42 +4,34 @@
...
-  return Buffer.from(signingKeyEncoded, 'base64');
-}
-const cachedSigningKey = memoize(signingKey);
-
 /*
  * Creates a standard Meadowlark Jwt.
  */
 function createToken(clientId: string, vendor: string, roles: string[]): Jwt {
-  const claims = { iss: 'ed-fi-meadowlark', aud: 'meadowlark', roles, client_id: clientId };
+  const claims = { iss: TOKEN_ISSUER, aud: TOKEN_ISSUER, roles, client_id: clientId };
```
stephenfuqua, line 32: These will be environment variables in a future ticket, FYI. Easy enough to rename when we get there - so no need to modify right now.

Meadowlark-js/packages/meadowlark-authz-server/src/handler/VerifyToken.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  let unvalidatedBody: any;
+
+  // startsWith accounts for possibility of the content-type being with or without encoding
+  if (!authorizationRequest.headers['content-type']?.startsWith('application/x-www-form-urlencoded')) {
+    Logger.debug(
+      `${moduleName}.parseVerifyTokenBody: Requires application/x-www-form-urlencoded content type`,
+      authorizationRequest.traceId,
+    );
+    return { isValid: false, failureMessage: 'Malformed body' };
```
stephenfuqua, line 46: Can we make the message more specific? "Invalid content-type" or something.

Meadowlark-js/packages/meadowlark-authz-server/src/handler/VerifyToken.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  try {
+    unvalidatedBody = querystring.parse(authorizationRequest.body);
+  } catch (error) {
+    Logger.debug(`${moduleName}.parseRequestTokenBody: Malformed body`, authorizationRequest.traceId);
+    return { isValid: false, failureMessage: 'Malformed body' };
```
stephenfuqua, line 53: Rather than having to rediscover this with an upcoming ticket on "malformed body", can you go ahead and report the error now? So long as we ensure the _content_ of the body is not part of the error message.

Meadowlark-js/packages/meadowlark-authz-server/src/model/SigningKey.ts
```diff
@@ -0,0 +1,15 @@
...
+
+function getSigningKey(): Buffer {
+  const signingKeyEncoded = process.env.SIGNING_KEY;
+  if (signingKeyEncoded == null) {
+    throw new Error('Must have a base-64 encoded signing key. Try creating a new one with `yarn createKey`');
+  }
+  return Buffer.from(signingKeyEncoded, 'base64');
+}
+
+export const signingKey = memoize(getSigningKey);
```
stephenfuqua, line 15: I've heard you mention memoize before, but it isn't really in my consciousness yet. I had to study this a bit to understand it. Should I have used this in the environment variable cache work?

Isn't the signing key needed on the core meadowlark API as well, for validating a token? So would it make sense to have this in a shared utility library, maybe in the Environment module? 

I agree about not using the existing `getStringFromEnvironment` function, because it is useful to have this custom error message. Or could add custom error message as an an optional argument on that method.

Meadowlark-js/packages/meadowlark-authz-server/src/security/TokenValidator.ts
```diff
@@ -0,0 +1,134 @@
...
+        sub: jwt.body?.sub ?? '',
+        aud: jwt.body?.aud ?? '',
+        iss: jwt.body?.iss ?? '',
+        exp: jwt.body?.exp ?? 0,
+        iat: jwt.body?.iat ?? 0,
+        roles: jwt.body?.roles ?? [],
+      },
+    };
+  } catch (err) {
+    Logger.debug('TokenValidator.introspectBearerToken user-submitted token error', traceId, err);
```
stephenfuqua, line 131: Are you sure that such an exception is just a user problem, not a problem with our software?

Meadowlark-js/packages/meadowlark-authz-server/test/security/TokenValidator.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      {
+        "errorResponse": {
+          "body": "{ "error": "invalid_client", "error_description": "Authorization token not provided" }",
+          "headers": {
+            "WWW-Authenticate": "Bearer",
+          },
+          "statusCode": 401,
+        },
+        "isValid": false,
+      }
```
stephenfuqua, line 58: That's a nice looking error response.

Meadowlark-js/services/meadowlark-fastify/src/Service.ts
```diff
@@ -119,9 +120,8 @@ export function buildService(): FastifyInstance {
     fastify.get(`/${stage}/metadata/descriptors/swagger.json`, swaggerForDescriptorsAPI);
     fastify.get(`/${stage}/metadata/data/v3/dependencies`, dependencies);
 
-    // OAuth handlers
+    // Old OAuth handlers
```
stephenfuqua, line 123: 🤨

Meadowlark-js/packages/meadowlark-authz-server/src/security/TokenValidator.ts
```diff
@@ -107,12 +107,14 @@ export async function introspectBearerToken(bearerToken: string, traceId: string
...
+    // Check for error response from nJwt
     if (jwt == null || nJwtErrorMessages.includes(jwt.message)) return { isValid: false };
 
-    const isNotExpired: boolean = jwt.message !== 'Jwt is expired';
-    const isMeadowlarkToken: boolean = jwt.body?.iss === TOKEN_ISSUER && jwt.body?.aud === TOKEN_ISSUER;
+    // Check for correct issuer
+    if (jwt.body?.iss !== TOKEN_ISSUER || jwt.body?.aud !== TOKEN_ISSUER) return { isValid: false };
 
-    const active: boolean = isNotExpired && isMeadowlarkToken && (await isValidClientId(jwt.body?.client_id, traceId));
+    const isNotExpired: boolean = jwt.message !== 'Jwt is expired';
```
stephenfuqua, line 116: I wondered where this text comes from. Answer: [a configurable file](https://github.com/jwtk/njwt/blob/master/properties.json).

This might not be worthwhile... wondering if we can do this?

```
import nJwtProperties from 'njwt/properties.json';

... 

const isNotExpired: boolean = jwt.message !==nJwtProperties.errors.EXPIRED;
```

Meadowlark-js/packages/meadowlark-authz-server/src/handler/VerifyToken.ts
bradbanister: 👍  Meant to, probably copy/paste error.

Meadowlark-js/packages/meadowlark-authz-server/src/model/SigningKey.ts
bradbanister: I used memoize because we transform the environment variable into a Buffer before providing it, so this just prevents doing that every time.
I think we are moving token validation exclusively to the auth server in RND-336.

Meadowlark-js/packages/meadowlark-authz-server/src/security/TokenValidator.ts
```diff
@@ -0,0 +1,134 @@
...
+      isValid: true,
+      introspectedToken: {
+        active,
+        client_id: jwt.body?.client_id ?? '',
+        sub: jwt.body?.sub ?? '',
+        aud: jwt.body?.aud ?? '',
+        iss: jwt.body?.iss ?? '',
+        exp: jwt.body?.exp ?? 0,
+        iat: jwt.body?.iat ?? 0,
+        roles: jwt.body?.roles ?? [],
```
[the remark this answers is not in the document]
bradbanister: This is the unpacked token provided by the user and unpacked by nJwt, so probably best to be defensive.

Meadowlark-js/packages/meadowlark-authz-server/src/security/TokenValidator.ts
bradbanister: Actually, it's neither! Turns out nJwt throws exceptions for not-very-exceptional conditions, like expired token. Fixed this.

Meadowlark-js/services/meadowlark-fastify/src/Service.ts
bradbanister: The old OAuthHandler in meadowlark-core is only 9 lines of code now.

Meadowlark-js/packages/meadowlark-authz-server/src/security/TokenValidator.ts
bradbanister: I'll take a look 👍

Meadowlark-js/packages/meadowlark-authz-server/src/handler/VerifyToken.ts
bradbanister: Done

Meadowlark-js/packages/meadowlark-authz-server/test/security/TokenValidator.test.ts
```diff
@@ -167,23 +167,13 @@ describe('given introspection of expired token', () => {
...
-        "isValid": true,
-      }
-    `);
+    expect(result.isValid).toBe(true);
+    expect((result as any).introspectedToken.active).toBe(false);
+    expect((result as any).introspectedToken.aud).toBe(TOKEN_ISSUER);
+    expect((result as any).introspectedToken.client_id).toBe('clientId');
+    expect((result as any).introspectedToken.iss).toBe(TOKEN_ISSUER);
+    expect((result as any).introspectedToken.roles).toHaveLength(1);
+    expect((result as any).introspectedToken.roles[0]).toBe('vendor');
```
stephenfuqua, line 176: I don't have any concern with this, as I am not used to snapshots anyway. Just curious: why change styles?

Meadowlark-js/packages/meadowlark-authz-server/src/security/TokenValidator.ts
bradbanister: A good idea, but we can't reach in there to the `properties.json` without file system hacking. It doesn't look exported.

Meadowlark-js/packages/meadowlark-authz-server/src/model/SigningKey.ts
stephenfuqua: That's what I get for reviewing during a meeting - forgetting the entire point of this! Treat Meadowlark API as a client of the OAuth server. It asks the OAuth server for verification, and it does not need to have the signing key.

Meadowlark-js/packages/meadowlark-authz-server/test/security/TokenValidator.test.ts
bradbanister: Because the token includes timestamps. Always a surprise when the snapshot looks good at first but won't pass the second time. 🤣

stephenfuqua (approved):

PR #124, opened 2022-10-21
[RND-428] Allow empty array for optional collections

stephenfuqua:
And opportunistically pin minimatch to fix a security alert


Meadowlark-js/tests/e2e/SchemaValidation.test.ts
```diff
@@ -0,0 +1,215 @@
...
+            },
+          ],
+        })
+        .expect(400)
+        .then((response) => {
+          expect(response.body.message).toMatchInlineSnapshot(`
+            [
+              " must have required property 'categories'",
+            ]
+          `);
```
stephenfuqua, line 34: @andonyns this is the first time I have used an InlineSnapshot with Jest in my own test. It was really easy - I just wrote expect(response.body.message).toMatchInlineSnapshot('');` and let the test fail. Then I let jest (via VS Code integration) update all of the snapshots in the test. 

Then I could review the messages and ensure they are what I actually expected.

When we improve the error messages, it will be simple to update all of these snapshots automatically, and again manually inspect to make sure they are what we want. In that case, if the update results in message we don't like, that means a) the code is actually outputing the wrong message, and b) we should manually correct the inline snapshot.

andonyns: I saw this on the system tests and I like this approach

bradbanister: Yes this is very handy. Jest also provides an excellent diff when a snapshot fails, making it easy to see what went wrong.

bradbanister (approved):
Looks really good

PR #126, opened 2022-10-24
[RND-365] Run End To End Tests

andonyns:
- Adding end to end tests running Fastify and Postgres Backend
- Separating testing process into two jobs
- Adding MongoDB Action to run Mongo Tests
- Including new files to reporter.

Must be approved after https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-Actions/pull/32 is merged.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,8 +73,16 @@ jobs:
           retention-days: 1
 
   end-to-end:
-    name: End to End tests
+    name: End to End tests for ${{ matrix.config.db }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          [
+            { db: "PGSQL", plugin: "@edfi/meadowlark-postgresql-backend" },
+            { db: "Mongo", plugin: "@edfi/meadowlark-mongodb-backend" },
+          ]
 
     defaults:
       run:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -71,3 +71,99 @@ jobs:
...
+
+      - name: Start OpenSearch Container
+        run: docker compose -f ./backends/meadowlark-opensearch-backend/docker/docker-compose.yml up -d
+
+      - if: ${{ matrix.config.db == 'PGSQL' }}
+        name: Start Postgres Container
+        run: docker compose -f ./backends/meadowlark-postgresql-backend/docker/docker-compose.yml up -d
+
+      - if: ${{ matrix.config.db == 'Mongo' }}
+        name: Start MongoDB
```
andonyns, line 113: Using MongoDB Action instead of setting up manual decreased the execution time from [7 minutes](https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/actions/runs/3349414681/jobs/5549393468) to [5 minutes](https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/actions/runs/3349518822/jobs/5549596165).

After seeing the time to setup Mongo, I don't think it's necessary to continue the effort to use in memory db

.github/workflows/test.yml
```diff
@@ -71,3 +71,99 @@ jobs:
...
+          retention-days: 10
+
+      # Workaround for issue https://github.com/actions/upload-artifact/issues/174
+      - name: Upload test results
+        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        with:
+          name: test-results
+          path: |
+            Meadowlark-js/${{matrix.config.db}}-e2e-tests/junit.xml
+            Meadowlark-js/.prettierrc
```
andonyns, line 168: The reporter issue causes that when a single file is attached, the folder structure is not preserved, to maintain the structure, it's necessary to attach any other file into the generated artifact.

eng/setup-mongo.sh
```diff
@@ -0,0 +1,22 @@
+cd ../Meadowlark-js/backends/meadowlark-mongodb-backend/docker/
```
andonyns, line 1: This script is currently not being used, but I consider it can be useful for getting started instructions to setup the local environment from scratch. I can leave it in the PR or add it in another one.

.github/workflows/test.yml
```diff
@@ -71,3 +71,99 @@ jobs:
...
+        with:
+          node-version: "16"
+          cache: "yarn"
+          cache-dependency-path: "**/yarn.lock"
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Start OpenSearch Container
+        run: docker compose -f ./backends/meadowlark-opensearch-backend/docker/docker-compose.yml up -d
```
stephenfuqua, line 106: In a future story, we should experiment with using [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) to see (a) if they have any performance benefits and (b) which use cases they might help with. For (b), I don't know, for example, if there is an available OpenSearch container that will run as a service easily.

Frankly, at some point I forgot about service containers, and the pattern seen here has been used in several places already. That's why I'm not concerned about changing it right now. Just need to investigate.

.github/workflows/test.yml
```diff
@@ -71,3 +71,99 @@ jobs:
...
+
+      - name: Cache Build
+        uses: actions/cache@48af2dc4a9e8278b89d7fa154b955c30c6aaab09 #v3.0.2
+        env:
+          cache-name: cache-build
+        with:
+          path: "**/dist/**"
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/dist/**') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
```
stephenfuqua, line 128: This caching is nice 😁

.github/workflows/test.yml
```diff
@@ -71,3 +71,99 @@ jobs:
...
+
+      - if: ${{ steps.cache-build.outputs.cache-hit != 'true' }}
+        name: Build
+        run: yarn build
+
+      - name: Run Fastify in background
+        run: |
+          cp ./tests/e2e/setup/e2e.env ./services/meadowlark-fastify/.env
+          yarn start:local > fastify.log &
```
stephenfuqua, line 137: I like this log redirection, so that the GitHub Actions output isn't overwhelmed, but we still have access to the detail if needed.

eng/setup-mongo.sh
stephenfuqua: Can leave it

stephenfuqua (approved):

PR #134, opened 2022-11-08
[RND-358] Authz server supports admin bootstrap

bradbanister:


Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/authorization/TryCreateBootstrapAuthorizationAdmin.ts
```diff
@@ -0,0 +1,57 @@
...
+        asUpsert(session),
+      );
+
+      if (!acknowledged) {
+        const msg =
+          'mongoCollection.replaceOne returned acknowledged: false, indicating a problem with write concern configuration';
+        Logger.error(functionName, request.traceId, msg);
+      } else {
+        Logger.debug(`${functionName}: Inserted ${upsertedCount} documents`, request.traceId);
+        createResult.response = upsertedCount === 1 ? 'CREATE_SUCCESS' : 'CREATE_FAILURE_ALREADY_EXISTS';
```
stephenfuqua, line 48: If the super user account has already been created, and someone tries to create another, that could be someone trying to attack the software. Might be worthwhile to get more visibility to that. I think we're quite safe, so WARN might be too strong. Perhaps it would make sense to log at the INFO level when someone tries to create a super user and one already exists.

stephenfuqua: I'm implementing this change.

stephenfuqua (approved):

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/authorization/TryCreateBootstrapAuthorizationAdmin.ts
```diff
@@ -44,7 +44,14 @@ export async function tryCreateBootstrapAuthorizationAdminDocument(
           'mongoCollection.replaceOne returned acknowledged: false, indicating a problem with write concern configuration';
         Logger.error(functionName, request.traceId, msg);
       } else {
-        Logger.debug(`${functionName}: Inserted ${upsertedCount} documents`, request.traceId);
+        if (upsertedCount == 1) {
```
andonyns, line 47: `if (upsertedCount === 1) {`

andonyns (changes requested):

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/authorization/TryCreateBootstrapAuthorizationAdmin.ts
andonyns: Not sure if the double equal comparison is intentional since I see other parts of the code where the verification is done with loose equality.

stephenfuqua: Should be triple equal - I commonly miss that when switching between Python (or C#) and TypeScript. Linter should be failing too.

stephenfuqua:
```suggestion
        if (upsertedCount === 1) {
```

andonyns (approved):

PR #135, opened 2022-11-08
[RND-416] Migrate Integration Test Suite

andonyns:
- Created tests for limit, offset and query parameters.
- Created readme file to list the existing scenarios and avoid repetition.
- Remove http test that no longer applies
- Created two folders for e2e tests: Scenarios and Endpoints, the goal is to separate the old tests that were previously created to determine if we want to use this tests to test a specific scenario, if not, eventually remove the code


Meadowlark-js/tests/e2e/functions/Resources.ts
```diff
@@ -19,8 +20,12 @@ export async function createResource({
     .post(`/v3.3b/ed-fi/${endpoint}`)
     .auth(await getAccessToken(credentials), { type: 'bearer' })
     .send(body)
-    .expect(201)
     .then((response) => {
+      if (response.body) {
+        console.error(response.body);
```
stephenfuqua, line 25: This seems reasonable in test code. Should we customize your slide not rules for these tests to stop flagging console access?

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
```diff
@@ -0,0 +1,287 @@
...
+      describe('when getting less than total', () => {
+        // Use the assessment1 credentials to bypass additional validations
+        it('should return the total and a subset of the results', async () => {
+          const limit = total - 1;
+          await baseURLRequest
+            .get(`/v3.3b/ed-fi/schools?limit=${limit}`)
+            .auth(await getAccessToken(Clients.Assessment1), { type: 'bearer' })
+            .expect(200)
+            .then((response) => {
+              expect(+response.headers['total-count']).toEqual(total);
```
stephenfuqua, line 47: What does that `+` mean?

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
```diff
@@ -0,0 +1,287 @@
...
+              expect(response.body.length).toEqual(limit);
+            });
+        });
+      });
+
+      describe('when requesting with invalid values', () => {
+        // Use the assessment1 credentials to bypass additional validations
+        it.each([0, 'zero'])('limit = %s', async (limit) => {
```
stephenfuqua, line 55: Can we add a few injection / attack inputs?

```suggestion
        it.each([0, 'zero', '5; select * from users', '0)', '1%27'])('limit = %s', async (limit) => {
```

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
bradbanister: I believe it forces a coercion to number. I'd prefer to do it explicitly.

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
```diff
@@ -0,0 +1,287 @@
...
+              expect(+response.headers['total-count']).toEqual(total);
+              expect(response.body.length).toEqual(0);
+              expect(response.body).toMatchInlineSnapshot(`[]`);
+            });
+        });
+      });
+
+      describe('when requesting with invalid values', () => {
+        // Use the assessment1 credentials to bypass additional validations
+        it.each([0, 'zero'])('offset = %s', async (offset) => {
```
stephenfuqua, line 109:
```suggestion
        it.each([0, 'zero', '5; select * from users', '0)', '1%27'])('offset = %s', async (offset) => {
```

Meadowlark-js/tests/.eslintrc
```diff
@@ -3,6 +3,7 @@
     "jest": true
   },
   "rules": {
+    "no-console": "off",
```
bradbanister, line 6: Can we move this down to an .eslintrc for the e2e directory only?

Meadowlark-js/tests/e2e/functions/BulkCreation.ts
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+import { createSchool } from './DataCreation';
+import { deleteResourceByLocation } from './Resources';
+
+class Resources {
```
bradbanister, line 9: Absent a compelling reason to be a class, this is better off as simply a type.

Meadowlark-js/tests/e2e/functions/BulkCreation.ts
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
```
bradbanister, line 5: Could we change the "functions" directory name to something like "helpers". It seems like these are all test helpers, and "functions" is pretty generic.

Meadowlark-js/tests/e2e/functions/BulkCreation.ts
bradbanister: More generally, note that while in C#-land classes also incorporate the concepts of types and modules, in TypeScript-land, they are all separable.

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
andonyns: I'll do the opposite and instead of converting the header, I will "stringify" the number

stephenfuqua (approved):

PR #138, opened 2022-11-11
[RND-427] - Unable to query school by schoolId

stevenarnold-ed-fi:
Fix query parameters not validating for non string values
Also fixed related unit tests that were being skipped Added unit tests for query parameter validation

### DIFF
--- a/Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
+++ b/Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
@@ -95,8 +95,8 @@ export function validateQueryParametersAgainstSchema(
    * If the value can't be converted (i.e. a word is provided for a numeric value, we leave it and let ajv invalidate
    * the schema).
    */
-  // eslint-disable-next-line no-restricted-syntax
-  for (const keyValue of parameterKeys) {
+
+  parameterKeys.forEach((keyValue) => {
     const property = schema.properties[keyValue];
 
     if (property != null && property.type != null) {
@@ -110,11 +110,13 @@ export function validateQueryParametersAgainstSchema(
         queryParameters[keyValue] = Boolean(queryParameters[keyValue]);
       }
     }
-  }
+  });
 
   const valid: ValidateFunction = ajv.compile(schema);
 
-  valid(queryParameters);
+  const isValid: boolean = valid(queryParameters);
+
+  if (isValid) return [];
 
   if (valid.errors) {
     const ajvErrors = valid.errors.map((error: ErrorObject) => {
--- a/Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
+++ b/Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
@@ -77,16 +77,57 @@ export function validateQueryParametersAgainstSchema(
 ): string[] {
   // TODO: RND-67 removed the Joi-specific code to report invalid query parameters
   // RND-307 will restore as a part of understanding Ajv error metadata
+
+  let errors: string[] = [];
+
   const schema = {
     ...metaEdModel.data.meadowlark.jsonSchema,
     // Need to relax the validation such that no fields are "required"
     required: [],
   };
 
+  const parameterKeys = Object.keys(queryParameters);
+
+  /**
+   * Number and boolean are come through as strings, which causes ajv to fail validation of parameters
+   * This loops through all query parameters and checks the type against the schema, if the type is numeric or boolean
+   * it will attempt convert the value. If the conversion is good, we set the value back and ajv will validate the schema
+   * If the value can't be converted (i.e. a word is provided for a numeric value, we leave it and let ajv invalidate
+   * the schema).
+   */
+  // eslint-disable-next-line no-restricted-syntax
+  for (const keyValue of parameterKeys) {
+    const property = schema.properties[keyValue];
+
+    if (property != null && property.type != null) {
+      if ((property && property?.type === 'integer') || property.type === 'number') {
+        const value = Number(queryParameters[keyValue]);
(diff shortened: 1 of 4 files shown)


Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
```diff
@@ -77,16 +77,57 @@ export function validateQueryParametersAgainstSchema(
...
+
+  /**
+   * Number and boolean are come through as strings, which causes ajv to fail validation of parameters
+   * This loops through all query parameters and checks the type against the schema, if the type is numeric or boolean
+   * it will attempt convert the value. If the conversion is good, we set the value back and ajv will validate the schema
+   * If the value can't be converted (i.e. a word is provided for a numeric value, we leave it and let ajv invalidate
+   * the schema).
+   */
+  // eslint-disable-next-line no-restricted-syntax
+  for (const keyValue of parameterKeys) {
```
stephenfuqua, line 99: Why not follow eslint's advice here? I don't remember the exact reason, but it is my understanding that the following is a safer approach (perhaps because of the possibility of `parameterKeys` being modified during enumeration):

```suggestion
  Object.keys(queryParameters).forEach((key) => {
```

Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
```diff
@@ -77,16 +77,57 @@ export function validateQueryParametersAgainstSchema(
...
+      if (property.type === 'boolean' && (queryParameters[keyValue] === 'true' || queryParameters[keyValue] === 'false')) {
+        queryParameters[keyValue] = Boolean(queryParameters[keyValue]);
+      }
+    }
+  }
+
   const valid: ValidateFunction = ajv.compile(schema);
 
-  const isValid: boolean = valid(queryParameters);
+  valid(queryParameters);
```
stephenfuqua, line 117: This is a strange object: `valid` is both a function call and it has properties (i.e. `errors`). Not much we can do about that.

I liked the older pattern of having `isValid` and `if (isValid) return [];` because it provides a "fast" response to the positive situation, and then the negative situation is at the root of the function rather than being indented. I just find it a little easier to read when the "quick path" returns first before the code path that needs to do more work to create the return value.

Meadowlark-js/packages/meadowlark-core/test/metaed/MetaEdValidation.test.ts
```diff
@@ -31,47 +48,230 @@ const createModel = (): TopLevelEntity => ({
...
+//             type: 'bool',
+//           },
+//         },
+//         required: [],
+//         title: 'EdFi.StudentEducationOrganizationAssociation',
+//         type: 'object',
+//       },
+//     },
+//   },
+// });
```
stephenfuqua, line 77: Can this code be deleted?

Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
bradbanister: Agree. We do occasionally override eslint on loops when the forEach function gets tangled up with async/await issues, but that's because using forEach would actually be wrong.

Meadowlark-js/packages/meadowlark-core/test/metaed/MetaEdValidation.test.ts
stevenarnold-ed-fi: Yep, accidental miss - deleted

Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
stevenarnold-ed-fi: This was a remnant from when I thought I was going to handle the errors from the validation code above myself, didn't need to and didn't revert back to this style, updated.

Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
stevenarnold-ed-fi: was from the data store code where we were making async calls during the foreach, which is bad. I took that a little too much to heart. No async calls in this one, so fixed.

Meadowlark-js/packages/meadowlark-core/src/metaed/MetaEdValidation.ts
```diff
@@ -77,16 +77,57 @@ export function validateQueryParametersAgainstSchema(
...
     required: [],
   };
 
+  const parameterKeys = Object.keys(queryParameters);
+
+  /**
+   * Number and boolean are come through as strings, which causes ajv to fail validation of parameters
+   * This loops through all query parameters and checks the type against the schema, if the type is numeric or boolean
+   * it will attempt convert the value. If the conversion is good, we set the value back and ajv will validate the schema
+   * If the value can't be converted (i.e. a word is provided for a numeric value, we leave it and let ajv invalidate
```
bradbanister, line 95: Are their cases where this could be a false-negative? Specifically where a number is provided in the query but is meant as a string for a string field?

stephenfuqua: This worked just fine: 
```
### 
POST http://localhost:3000/local/v3.3b/ed-fi/students
authorization: bearer {{authToken4}}

{ 
    "studentUniqueId": "1",
    "firstName": "Hello",
    "lastSurname": "World",
    "birthDate": "2001-01-01",
    "birthCountryDescriptor": "uri://ed-fi.org/CountryDescriptor#AD3"
}

###
GET http://localhost:3000/local/v3.3b/ed-fi/students?studentUniqueId=1
authorization: bearer {{authToken4}}
```

stephenfuqua (approved):

PR #141, opened 2022-11-14
[RND-439] Updating e2e tests to handle its own credentials

andonyns:


stephenfuqua (commented):
Looks pretty good to me. No concerns. Look forward to seeing everything in Action.

PR #147, opened 2022-11-16
[RND-439] Remove hardcoded credentials for e2e tests

andonyns:
- Remove hardcoded users (leave admin and verifyOnly).
- Add globalSetup function for e2e tests to set up admin user.
- Add function to set up clients required for testing.
- These values can be provided through environment values.


.github/workflows/test.yml
```diff
@@ -123,8 +123,7 @@ jobs:
           sudo systemctl start postgresql.service
           sudo -u postgres psql -U postgres -d postgres -c "alter user postgres with password '${{ env.POSTGRES_PASSWORD }}';"
 
-      - if: ${{ matrix.config.db == 'Mongo' }}
-        name: Start MongoDB
+      - name: Start MongoDB
```
stephenfuqua, line 126:
```suggestion
      # Always start MongoDB, as it is used by the authorization system.
      - name: Start MongoDB
```

Meadowlark-js/tests/e2e/helpers/Credentials.ts
```diff
@@ -3,74 +3,166 @@
...
+  key?: string;
+  secret?: string;
+  token?: string;
+};
+
+async function getAdminAccessToken(): Promise<string> {
+  let adminAccessToken = process.env.ADMIN_AT;
+  if (adminAccessToken == null) {
+    adminAccessToken = await baseURLRequest()
+      .post('/oauth/token')
```
stephenfuqua, line 23: I might modify this in my ticket, so that it reads this from the environment variables. But the env var will have a full URL, not just an endpoint - because of the possibility of integrating a different OAuth provider. Once we get to that point, we would probably still be running e2e tests using the built-in provider. So on second thought, I'll leave this in place. Keeping my comment though just to share some thought process.

Meadowlark-js/tests/e2e/helpers/Shared.ts
```diff
@@ -3,8 +3,23 @@
...
-import { Clients, getAccessToken } from './Credentials';
+import Chance from 'chance';
+import memoize from 'fast-memoize';
+import request, { SuperTest, Test } from 'supertest';
+import { getAccessToken } from './Credentials';
+
+const chance = new Chance() as Chance.Chance;
+
+function getBaseURLRequest() {
+  return request(process.env.BASE_URL);
```
stephenfuqua, line 14: `BASE_URL` appears to just be the root URL + stage. Could change BASE_URL to stage then, to more closely match fastify settings. But this is working, so probably not worth changing that right now.

andonyns: The only aspect to taken into account when we do this change is that the root URL is used to access resources by Location, because when a new resource is created, we use the location to delete the element after the test. The location path includes the stage, so this could be handled with logic

stephenfuqua (approved):

PR #148, opened 2022-11-18
[RND-355] Get Client By Id

stephenfuqua:


Meadowlark-js/packages/meadowlark-authz-server/src/message/GetAuthorizationClientResult.ts
```diff
@@ -11,6 +11,7 @@ export type GetAuthorizationClientResult =
       clientSecretHashed: string;
       clientName: string;
       roles: AuthorizationClientRole[];
+      active: boolean | undefined;
```
bradbanister, line 14: Why allow undefined?

bradbanister (approved):

Meadowlark-js/packages/meadowlark-authz-server/src/message/GetAuthorizationClientResult.ts
stephenfuqua: Temporary only. Meant to put a comment. Was trying to avoid breaking many things with this addition, since @stevenarnold-ed-fi's PR will have the real "active" work. Steven, please change to disallow undefined when you get to this point.

I only included it so that I could have it in the response body

PR #155, opened 2022-12-02
[RND-450] Env var to enable hard-coded credentials

stephenfuqua:


Meadowlark-js/packages/meadowlark-authz-server/src/handler/RequestToken.ts
```diff
@@ -169,8 +179,11 @@ export async function requestToken(authorizationRequest: AuthorizationRequest):
       }
 
       if (result.response === 'GET_FAILURE_NOT_EXISTS') {
-        Logger.debug(`${moduleName}.requestToken: ${maskClientSecret(requestTokenBody)} 404`, authorizationRequest.traceId);
-        return { body: '', statusCode: 404 };
+        Logger.debug(
+          `${moduleName}.requestToken: ${maskClientSecret(requestTokenBody)} Client does not exist`,
+          authorizationRequest.traceId,
+        );
+        return { body: '', statusCode: 401 };
```
stephenfuqua, line 186: 404 is not an appropriate response, because it would mean that the token endpoint doesn't exist. In this case, the authentication attempt is simply rejected because of invalid credentials.

Meadowlark-js/packages/meadowlark-utilities/src/Environment.ts
```diff
@@ -61,7 +62,7 @@ export function getValueFromEnvironment<T>(
 ): T {
   const cached = getFromCache<T>(key);
 
-  if (cached) {
+  if (cached !== undefined) {
```
stephenfuqua, line 65: Found a bug. If the cached value were the boolean `false`, then `if (cached)` was false, when it should be true.

andonyns (approved):

PR #161, opened 2022-12-06
[RND-452] - Update assessment role to act as vendor role

stevenarnold-ed-fi:


Meadowlark-js/tests/e2e/scenarios/AuthorizationValidation.test.ts
```diff
@@ -96,15 +96,9 @@ describe("given it's managing the client authorization", () => {
...
-        describe('when using an invalid combination of roles', () => {
-          it.each([
-            { roles: ['vendor', 'vendor'] },
-            { roles: ['vendor', 'host'] },
-            { roles: ['admin', 'host'] },
-            { roles: ['admin', 'vendor'] },
-            { roles: ['verify-only', 'host'] },
-          ])('create client with %j', async (item) => {
-            const { roles } = item;
+        describe('when generating a client with more than 2 roles', () => {
```
stephenfuqua, line 99: This causes me to wonder if there is now any reason to allow 2 roles? If "assessment" is like "vendor", only more permissive, then there is no need to have both.

The ticket says we decided to allow all combinations. But now I'm wondering if _any_ combination makes sense, in the context of treating assessment as a special case of vendor?

stephenfuqua: On the other hand, the whole point of having plural "roles" instead of "role" is to allow multiple things.

stevenarnold-ed-fi: On the first comment, I had the same thought, but went with the ticket description. Happy to discuss during standup and adjust if needed

stephenfuqua (approved):

PR #163, opened 2022-12-08
[RND-402] Standardize the logging

stephenfuqua:


Meadowlark-js/packages/meadowlark-core/src/handler/Update.ts
```diff
@@ -72,7 +72,7 @@ export async function update(frontendRequest: FrontendRequest): Promise<Frontend
       };
     }
 
-    writeDebugStatusToLog(moduleName, frontendRequest, 'update', 500, failureMessage);
+    writeErrorToLog(moduleName, frontendRequest.traceId, 'update', 500, failureMessage);
```
andonyns, line 75: Lines 62 and 67should also be errors instead of debug messages (I wasn't able to add a suggestion for those lines)

andonyns (changes requested):
Confirm if suggested messages should also be errors instead of debug

Meadowlark-js/packages/meadowlark-core/src/handler/Update.ts
stephenfuqua: I agree about this specific line - it should be an error, because we don't really know what happened, and had to respond with 500. But the other two cases should remain as debug messages, because these are _user_ errors instead of a _application_ error. The ERROR log level is intended for system administrators to diagnose how the application is running, and it should only be used for _application errors_.

andonyns (approved):

PR #165, opened 2022-12-12
[RND-403] - Anonymized Request Body Logging

stevenarnold-ed-fi:
image

Wasn't sure if we wanted the standard writeRequestToLog call in this code, so included it for now, can remove if unneeded


Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
```diff
@@ -0,0 +1,31 @@
...
+import { MiddlewareModel } from './MiddlewareModel';
+
+const moduleName = 'core.middleware.ParseBodyMiddleware';
+
+export async function anonymizeAndLogRequestBody({
+  frontendRequest,
+  frontendResponse,
+}: MiddlewareModel): Promise<MiddlewareModel> {
+  // if there is a response already posted, we are done
+  if (frontendResponse != null) return { frontendRequest, frontendResponse };
```
stephenfuqua, line 17: How would we ever fall through to line 19?

Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (isDebugEnabled() && !frontendRequest.middleware.resourceInfo.isDescriptor) {
+    writeRequestToLog(moduleName, frontendRequest, 'anonymizeAndLogRequestBody');
```
stephenfuqua, line 20: In this case, we're about to write a debug message. Don't think we need to also write an info message indicating that we entered the block of code.

Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const copyOfParsedBody = { ...frontendRequest.middleware.parsedBody };
+    Object.keys(copyOfParsedBody).forEach((key) => {
+      copyOfParsedBody[key] = null;
+    });
+
+    Logger.debug(`Anonymized request body: ${JSON.stringify(copyOfParsedBody)}`, frontendRequest.traceId);
```
stephenfuqua, line 27: It would be nice to log `copyOfParsedBody` as an object instead of stringified. There should be an optional argument that supports that.

Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
stevenarnold-ed-fi: The frontend response is generally null at this point, if there is already a response we don't continue processing

Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
stevenarnold-ed-fi: Just following standard practice but had the same thought, and agree it's not needed, will remove

Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
stephenfuqua: I got myself mixed up between request and response. This makes sense to me now, when I read it correctly.

Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
stevenarnold-ed-fi: The logger in Meadowlark utilities only exposes string messages

stevenarnold-ed-fi: @stephenfuqua logging updated

stephenfuqua:
This PR looks great to me but I haven't been able to run it yet locally, so not marking as approved.

Meadowlark-js/packages/meadowlark-core/src/middleware/LogAnonymizedRequestMiddleware.ts
```diff
@@ -0,0 +1,26 @@
...
+import { MiddlewareModel } from './MiddlewareModel';
+
+export async function anonymizeAndLogRequestBody({
+  frontendRequest,
+  frontendResponse,
+}: MiddlewareModel): Promise<MiddlewareModel> {
+  // if there is a response already posted, we are done
+  if (frontendResponse != null) return { frontendRequest, frontendResponse };
+
+  if (isDebugEnabled() && !frontendRequest.middleware.resourceInfo.isDescriptor) {
```
stephenfuqua, line 16: The original idea is that descriptors _inside the body_ do not need to be anonymized. This code _is_ anonymizing them. I now think that's okay, because the client will get a 400 response that tells them the descriptor doesn't exist, and that response can be logged too.

For debugging purposes, it will still be handy to see what the payload structure is for a descriptor, therefore let's remove this second condition.
```suggestion
  if (isDebugEnabled()) {
```

stephenfuqua: Ooh... here's a bigger challenge: what about nested documents?

```
POST http://localhost:3000/local/v3.3b/ed-fi/StudentInterventionAssociations
authorization: bearer {{authToken4}}

{ 
    "studentReference": {
        "studentUniqueId": "s0zf6d1123d3e"
    },
    "interventionReference": {
        "interventionIdentificationCode": "111",
        "educationOrganizationId": 123
    }
}
```

-->

```
2023-01-04 14:37:74 debug 3ffbd49e-2229-4f1e-aad2-e952c2e46f37 Anonymized request body: {"studentReference":null,"interventionReference":null}
```

stephenfuqua: @stevenarnold-ed-fi I created a [pull request into your branch](https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/pull/177), please take a look.

stephenfuqua (approved):

PR #166, opened 2022-12-15
[RND-413] Add test for multiple references

andonyns:


Meadowlark-js/tests/e2e/helpers/Resources.ts
```diff
@@ -21,7 +21,7 @@ export async function createResource({
     .send(body)
     .then((response) => {
       if (response.body) {
-        console.error(response.body);
+        console.error(`Error on ${endpoint}:\n${response.body}`);
```
stephenfuqua, line 24: Should we have an `expect(response.body).toBeUndefined()` or something like that here, so that the test fails in this situation?

andonyns: I like this idea

stephenfuqua (approved):

stephenfuqua:
@andonyns somehow you have a couple of commits that are not signed, thus I'm unable to merge. You'll need to sign those locally and force push an update.

stephenfuqua (changes requested):
Retracting my approval because of the unsigned commits.

andonyns:
Interesting. I'll check

bradbanister (approved):

PR #168, opened 2022-12-16
[RND-462] Change structure of GitHub actions to create workflows per trigger action

andonyns:


.github/workflows/on-pullrequest.yml
```diff
@@ -1,10 +1,15 @@
-name: Test
+name: On Pull Request
 
 on:
   push:
-    branches: [main]
+    branches:
```
andonyns, line 5: For Meadowlark, we run the tests once the PR is open and after it's merged. This is contradictory with the name, but I consider this is a good practice. If required, we can remove this trigger to leave only as "On Pull Request".

.github/workflows/on-pullrequest.yml
```diff
@@ -16,6 +21,67 @@ defaults:
     working-directory: Meadowlark-js
 
 jobs:
+  scan-actions:
+    name: Scan Actions
+    uses: ed-fi-alliance-oss/ed-fi-actions/.github/workflows/repository-scanner.yml@latest
+
+  analyze-code:
+    if: github.event_name == 'pull_request'
```
andonyns, line 29: The dependency review should be executed only when the trigger is a PR

.github/workflows/on-pullrequest.yml
```diff
@@ -16,6 +21,67 @@ defaults:
...
+    name: Analyze Code Dependencies
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
```
stephenfuqua, line 39: Can you look to see if there are newer & better versions available for

* checkout
* dependency-review-action
* setup-node
* actions-cache
* github/codeql-action

.github/workflows/on-pullrequest.yml
andonyns: I was thinking of an scenario where a PR is opened, and then stays open for a couple of weeks while there are other PRs getting merged but since we don't allow automatic merges I don't think this scenario is feasible.

I don't think this adds any value unless scheduled or something similar, given that even if something changes on main, the workflow will not be executed unless there's a commit, which is not allowed. 

Therefore, I don't think we need this push on main

stephenfuqua: In the daily standup we decided that there is a very small, but real, risk of something going wrong during a merge, and therefore it will be better to go ahead and test on merge.

stephenfuqua (approved):

PR #169, opened 2022-12-16
[RND-404] Anonymized Response Logging

stephenfuqua:
Includes substantial refactor so that response bodies are passed around as object instead of strings.


Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Db.ts
```diff
@@ -63,6 +63,11 @@ export async function createConnectionPoolAndReturnClient(): Promise<PoolClient>
     // database, and disconnect. From there reconnect the pool to the meadowlark database and continue
 
     singletonDbPool = new Pool(dbConfiguration);
+
+    singletonDbPool.on('error', (err, _client) => {
+      Logger.error(`${moduleName} a PostgreSQL connection error occurred.`, null, err);
+    });
+
```
stephenfuqua, line 70: I didn't realize I had committed this. Came in while I was exploring this problem with Andony earlier in the week. Wonder if this is actually a better  place to put this, before the `connect` command instead of after it? Let's see what happens when the e2e tests run in GitHub.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Update.ts
```diff
@@ -59,7 +59,7 @@ export async function updateDocumentById(
         Logger.debug(`${moduleName}.updateDocument: Inserting document id ${id} failed due to invalid references`, traceId);
         updateResult = {
           response: 'UPDATE_FAILURE_REFERENCE',
-          failureMessage: `Reference validation failed: ${failures.join(',')}`,
+          failureMessage: { error: { message: 'Reference validation failed', failures } },
```
stephenfuqua, line 62: @bradbanister @andonyns the weird indexing issue we explored this morning was due to a previous iteration of the line above, where I dereferenced like this:

```failureMessage: { error: { message: 'Reference validation failed', ...failures } },```

File: Meadowlark-js/packages/meadowlark-core/src/validation/DocumentValidator.ts [hunk 3]
```diff
@@ -11,7 +11,7 @@ import { validateEntityBodyAgainstSchema } from '../metaed/MetaEdValidation';
  *
  * Validates the request body for the resource. If invalid, returns an error message.
  */
-export async function validateDocument(body: object, matchingMetaEdModel: TopLevelEntity): Promise<string> {
+export async function validateDocument(body: object, matchingMetaEdModel: TopLevelEntity): Promise<object | null> {
```
bradbanister, line 14: Should be able to get more specific than `object`. How about defining `type ValidationFailure = { error: ValidationError[] };`

bradbanister (approved):
Looks good. One suggestion for improving type annotation, but approving.

PR #178, opened 2023-01-04
Configurable thread count

stephenfuqua:


Meadowlark-js/services/meadowlark-fastify/src/Cluster.ts
```diff
@@ -17,17 +17,18 @@ export class ClusterService {
...
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
       this.worker();
     }
   }
 
-  // eslint-disable-next-line class-methods-use-this
-  master() {
-    Logger.debug(`Total Number of Cores: ${CPUS}`, null);
-    Logger.debug(`Master ${process.pid} is running`, null);
+  static primary(): void {
```
stephenfuqua, line 26: Changed name to avoid the term "master", and made it static since it doesn't reference `this`.

NOTICES.md
```diff
@@ -15,3 +15,30 @@ distributed in its original form without modification.
 
 Please run the command `yarn licenses list` for a full list of all of the
 licenses used by this package's dependencies.
+
+--
```
stephenfuqua, line 19: I happened to find the code that was used to create this clustering. We need to give credit where credit is due when re-using code.

Meadowlark-js/services/meadowlark-fastify/src/Cluster.ts
bradbanister: This is fine but if it were me, I'd just take the function out of the class.

bradbanister (approved):
I like it

NOTICES.md
andonyns: I agree that this had to be done. Thanks for adding this

PR #182, opened 2023-01-18
[RND-456] Schema Correction for CohortYears array

stephenfuqua:
Note that there are a couple of new end-to-end tests that address the core issue, and also test other uses of SchoolYear. The second scenario ensures that the fix for the core problem (cohort year) does not cause problems for the other SchoolYear cases.


Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/UpdateOpensearch.ts
```diff
@@ -38,8 +65,7 @@ export async function afterDeleteDocumentById(request: DeleteRequest, result: De
     );
     await client.delete({ ...opensearchRequest, refresh: true });
   } catch (err) {
-    Logger.error(`${moduleName}.afterDeleteDocumentById`, request.traceId, err);
-    throw err;
+    throw safelyLogOpenSearchErrors(err);
```
stephenfuqua, line 68: This is difficult to test. I will see if I can come up with a manual test.

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/src/enhancer/JsonSchemaEnhancer.ts
```diff
@@ -258,7 +260,12 @@ function schemaPropertyForNonReference(property: EntityProperty): SchemaProperty
       return { type: 'string', format: 'time', description };
 
     case 'schoolYearEnumeration':
-      return { type: 'integer', description, minimum: beginYear, maximum: endYear };
+      if (property.parentEntity.type === 'common') {
+        // For a common, the school year ends up being nested under a reference object
+        return schoolYearEnumerationSchema;
+      }
+
+      return schoolYear;
```
stephenfuqua, line 268: This is the key change for the original problem. For a common, we end with something like

```
"cohortYears": [
    {
      "schoolYearTypeReference": { "schoolYear": 2022 },
      "cohortYearTypeDescriptor": "uri://ed-fi.org/CohortYearTypeDescriptor#Twelfth grade"
    }
```

But in other cases when we're in this function, we have a flatter structure like this:

```
"gradingPeriods": [
    {
      "gradingPeriodReference": {
        ...
        "schoolYear": 0
        }
      }
```

Note how `schoolYear` is flat here, instead of an object. So we need to support both possibilities.

Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/UpdateOpensearch.ts
stephenfuqua: I haven't been able to figure anything out yet. I think this code is written in a very safe way - but it needs careful visual scrutiny.

Meadowlark-js/package.json
```diff
@@ -55,7 +55,8 @@
     "test:system:mongodb": "cross-env DOCUMENT_STORE_PLUGIN=@edfi/meadowlark-mongodb-backend LOG_LEVEL=warn npx jest system --runInBand",
     "test:system:postgresql": "cross-env DOCUMENT_STORE_PLUGIN=@edfi/meadowlark-postgresql-backend LOG_LEVEL=warn npx jest system --runInBand",
     "test:unit:coverage": "rimraf coverage/ && npm run test:unit -- --collectCoverage",
-    "prepare": "cd .. && husky install"
+    "prepare": "cd .. && husky install",
+    "start:local": "lerna run start:local"
```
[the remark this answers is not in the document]
andonyns: Good to know. When we did the migration to npm I updated [LOCALHOST.md](https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/main/docs/LOCALHOST.md) with an extra step. With this fix we can remove that extra step

docs/LOCALHOST.md
```diff
@@ -48,6 +49,12 @@ POST */_delete_by_query
 }
 ```
 
+Delete all indices:
+
+```none
+DELETE *
+```
+
```
stephenfuqua, line 57: This and the readme in the OpenSearch docker directory have redundant instructions for clearing out the database. This is because it is easy to forget where the instructions are. If you are starting out for the first time, you're likely to look here. But if you're opening VS Code from `meadowlark-js` then you won't see this `docs` directory, and you'll be looking for the instructions somewhere else. Hence doubling up seemed useful in this circumstance.

Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/UpdateOpensearch.ts
```diff
@@ -22,6 +22,33 @@ const moduleName = 'opensearch.repository.UpdateOpensearch';
...
+    type: string;
+    reason: string;
+    caused_by: [{ type: string; reason: string }];
+  };
+};
+
+// Exported for the sake of unit testing
+export function safelyLogOpenSearchErrors(err: MapperParsingException | object): object {
+  const mapperParsingException = err as MapperParsingException;
+  if (mapperParsingException != null) {
```
bradbanister, line 37: Not sure I understand how this function is dealing with types. If `mapperParsingException` can be null, that can only be because `err` can be null, which isn't in the parameter type signature. Note that line 36 is `const mapperParsingException = err` at runtime, as Typescript type information is stripped out.

If it's not guaranteed that the shape of the thrown exception will match the MapperParsingException, you'll have to check that each property you want to use is not undefined.

Separately, if you know that the thrown object is at least an `Error`, that would be preferable to type `object`.

bradbanister (changes requested):

Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/UpdateOpensearch.ts
stephenfuqua: Reverse order: `Error` is easy enough in the signature.

At some point I suspected I was getting confused between TypeScript and C# again with the casting. Got distracted and forgot about the uncertainty.

I was looking for the easy way of "if the object contains this keyA, and object[keyA] contains keyB" etc.

stephenfuqua: I'm going to remove this code entirely, since we don't have a clear path for testing it. There is a backlog ticket for making OpenSearch more robust. We need to seriously investigate that, and need to figure out how to force some errors so that we can understand the structure. Looking at the source code from opensearch-js, for example, I don't even see a `root_cause` is defined as a string in TypeScript - but I saw an error message that had it as an array. So there are weird things going on. I'll make a note in that other ticket that we need to watch for potential PII in error messages.

Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/UpdateOpensearch.ts
stephenfuqua: I'm completely rolling back changes to this file. Other backend modules also have this double logging problem, and I'll create a new ticket to address that.

.github/workflows/on-pullrequest.yml
```diff
@@ -267,12 +272,15 @@ jobs:
...
+      - name: Restore build cache
         id: build-cache
-        uses: actions/cache@9b0c1fce7a93df8e3bb8926b0d6e9d89e92f20a7 #v3.0.11
+        uses: actions/cache/restore@58c146cc91c5b9e778e71775dfe9bf1442ad9a12 #v3.2.3
         with:
           path: "**/dist/**"
-          key: ${{ runner.os }}-build-${{ hashFiles('**/dist/**') }}
+          key: ${{ needs.build.outputs.hashKey }}
 
       - name: Build
```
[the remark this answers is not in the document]
stephenfuqua: You're right, it shouldn't happen. If it does happen, we have a safeguard - but probably something else will go wrong. I'm slightly in favor of leaving it in place though.

andonyns (approved):

bradbanister (approved):

PR #184, opened 2023-01-19
[RND-435] Refactor configuration settings

stephenfuqua:


.github/workflows/on-pullrequest.yml
```diff
@@ -208,6 +207,24 @@ jobs:
           sudo systemctl start postgresql.service
           sudo -u postgres psql -U postgres -c "alter user postgres with password '${{env.POSTGRES_PASSWORD}}';"
 
+      - name: Create .env file
```
andonyns, line 210: I think that with this setup, we could stop requiring the e2e.env file in line #305 and move this to a common place in the workflow. I haven't reviewed both files line by line to see if it covers all required values

andonyns, line 210: With these changes, could we remove the e2e.env file for e2e tests and use this same step if we move to a common place? I think this is doing for integration tests what e2e.env did for e2e tests

bradbanister (approved):

Meadowlark-js/backends/meadowlark-opensearch-backend/package.json
```diff
@@ -27,6 +27,7 @@
   "devDependencies": {
     "@short.io/opensearch-mock": "^0.3.1",
     "copyfiles": "^2.4.1",
+    "dotenv": "^16.0.3",
```
andonyns, line 30: The dotenv dependency could be in the root now given that all packages are using it, currently, we have 5 dependencies of dotenv and we've seen in the past that this could cause difference on versions through packages

PR #194, opened 2023-02-01
[RND-384] Increase test coverage for meadowlark-opensearch-backend folder

andonyns:
Add integration tests for Open Search

## Description

Add setup to run OpenSearch in TestContainers to be able to execute integration tests against it.
Add integration tests for OpenSearch backend. 

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have added tests to cover my changes.


Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
```diff
@@ -0,0 +1,133 @@
...
+  PaginationParameters,
+  QueryRequest,
+} from '@edfi/meadowlark-core';
+import { Client } from '@opensearch-project/opensearch/.';
+import { getNewClient } from '../../src/repository/Db';
+import { queryDocuments } from '../../src/repository/QueryOpensearch';
+import { afterUpsertDocument, afterDeleteDocumentById } from '../../src/repository/UpdateOpensearch';
+import { setupOpenSearch, teardownOpenSearch } from '../setup/OpenSearchSetup';
+
+jest.setTimeout(120_000);
```
stephenfuqua, line 23: An underscore?

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        await afterDeleteDocumentById(
+          { id, resourceInfo } as DeleteRequest,
+          { response: 'DELETE_SUCCESS' } as DeleteResult,
+          client,
+        );
+      }
+    });
+
+    it('should be able to retrieve the document', async () => {
+      const response = await queryDocuments(setupQueryRequest({}, {}), client);
```
stephenfuqua, line 88: What do you think about adding the following scenario?

Maybe try loading two students, one with firstName = 'Test' and one with firstName = 'Other'

Then do setup calls equivalent to the following GET requests:

* `GET students`  -> returns both
* `GET students/?Limit=1 ` -> returns only the first student uploaded (well, I'm not sure which one it will be, though it should always be the same on any system because of the `ORDER BY _doc` clause).
* `GET students/?Limit=1&Offset=1` -> returns the other student.
* `GET students/?firstName=Test` -> returns only that specific student
* `GET students/?firstName=Test&Offset=1` -> returns an empty array, since there is only one student to match
* `GET students/?firstName=\'injection+attack` -> returns nothing (I think), because we will end up searching for a blank first name instead.

Meadowlark-js/backends/meadowlark-opensearch-backend/test/setup/OpenSearchSetup.ts
```diff
@@ -0,0 +1,36 @@
...
+import path from 'path';
+import { DockerComposeEnvironment, StartedDockerComposeEnvironment, StartedTestContainer } from 'testcontainers';
+
+let environment: StartedDockerComposeEnvironment;
+const containerName = 'opensearch-test';
+
+export async function setupOpenSearch() {
+  let container: StartedTestContainer;
+  try {
+    const port = 8200;
```
stephenfuqua, line 16: Just an observation, no action: we are running our default solution on port 9200, so this should be fine.

Meadowlark-js/backends/meadowlark-opensearch-backend/test/setup/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3'
+services:
+
+  opensearch-node-integration:
+    image: opensearchproject/opensearch:latest
+    container_name: opensearch-test
```
stephenfuqua, line 6:
```suggestion
    container_name: opensearch-integration-test
```

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
andonyns: Yes, this is just "syntactic sugar" to express a bigger number. I left it the same as the value on the setup open search, but we could decrease it if needed, it does need a bigger value than the current default of 40000, but we could try and see to avoid intermittent errors

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
andonyns: I agree, this is needed. I will add this scenario

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
stephenfuqua: But what does it mean? Having a longer time is fine, but I have no clue how to interpret that 😆 .

andonyns: Got it. It's 120000

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
```diff
@@ -0,0 +1,305 @@
...
+
+        expect(result.response).toEqual('QUERY_SUCCESS');
+        expect(result.totalCount).toEqual(1);
+        expect(result.documents[0]).toEqual(student1);
+      });
+    });
+
+    describe('when querying with limit', () => {
+      // This is returning all values
+      it.skip('should return value', async () => {
```
andonyns, line 179: I added these tests but I'm skipping them until reviewing why it is returning all values despite filters

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
```diff
@@ -0,0 +1,305 @@
...
+
+        expect(result.response).toEqual('QUERY_SUCCESS');
+        expect(result.totalCount).toEqual(2);
+        expect(result.documents).toHaveLength(1);
+        expect(result.documents[0]).toEqual(student2);
+      });
+    });
+
+    describe('when querying with parameters and offset', () => {
+      it.skip('should return value', async () => {
```
andonyns, line 200: @stephenfuqua this scenario:
`GET students/?firstName=Test&Offset=1 -> returns an empty array, since there is only one student to match`
Is returning UNKNOWN_FAILURE instead of the mentioned result.
Is this the correct behavior?

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
```diff
@@ -0,0 +1,304 @@
...
+        expect(result.totalCount).toEqual(2);
+        expect(result.documents).toHaveLength(1);
+        expect(result.documents[0]).toEqual(student2);
+      });
+    });
+
+    describe('when querying with parameters and offset', () => {
+      it('should return value', async () => {
+        const result = await queryDocuments(
+          setupQueryRequest({ firstName: student1.firstName }, { limit: 2, offset: 1 }),
```
andonyns, line 202: @stephenfuqua this scenario was failing because we need to specify limit when using offset. This is not being limited in the [class](https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/main/Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/QueryOpensearch.ts#L100). I wonder if we should add a validation to make sure that both are specified.

If not present it will return UNKNOWN_FAILURE

stephenfuqua: We do have validation on that - its just that the validation is in middleware, not back here in the repository.

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/QueryOpensearch.test.ts
stephenfuqua: Agreed, this is an invalid scenario with OpenSearch, and I simply forgot about that. Can always use Limit=1, or Limit=10, etc. Point is, we are looking for a second entry and there is none.

stephenfuqua:
image

They're running! Wow, took a long time to run though. Might want to tune that in the future. Good enough for right now though.

stephenfuqua (approved):

PR #196, opened 2023-02-03
[RND-478] - Improves CRUD coverage in e2e tests.

DavidJGapCR:


Meadowlark-js/tests/e2e/scenarios/ResourcesGetValidation.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      .auth(await getAccessToken('host'), { type: 'bearer' })
+      .expect(200)
+      .then((response) => {
+        expect(response.body).toEqual(expect.objectContaining(educationContentBody));
+      });
+  });
+
+  afterAll(async () => {
+    await deleteResourceByLocation(educationContentLocation);
+    await deleteResourceByLocation(contentClassDescriptorLocation);
```
stephenfuqua, line 50: Note how the `ResourceGetValidation` test is also testing _delete_ functionality! and Of course it tested the `post` functionality.

We could probably have a single CRUD scenario like this:
```
describe when performing crud operations
  let location;

  describe when creating a new resource
    let getResponse;

    beforeAll
        location = ...
        // perform the post here and set `getResponse` in the `then` handler

    it returns 201
        expect(getResponse.status).toBe(201);


  describe when getting a resource by ID
    describe given the resource exists
      // Do a GET request with the location

    describe given the resource doe not exist
      // Do a GET request with a modified version of the location - for example, change the last character in the location string

  describe when getting all resources
    // expect 1 resource

  describe when upserting a resource
      // Run the post again and make sure it returns 200

  describe when updating a resource
      // Run the put with at least one property modified, and make sure it returns 204
      // Then run another GET by ID and make sure the value changed appropriately

  describe when  deleting a resource
      // Run the delete and make sure it returns 204
      // Then run another GET by ID and make sure it return 404
```

Meadowlark-js/tests/e2e/scenarios/SecurityValidation.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        .expect(201)
+        .then((response) => {
+          expect(response).not.toBe('');
+          educationContentLocation = response.headers.location;
+        });
+
+      expect(educationContentLocation).toBeTruthy();
+    });
+
+    it('Vendor 1 can get it', async () => {
```
stephenfuqua, line 113: Back to "given, when, then" - here's a case where it does make sense to keep "when" in between "given" and "then" (it) - or maybe just omit it entirely if it isn't helpful.

You could have something like this:

```
describe('when the same vendor requests the resource'
    it('should receive the resource')

describe('when a different vendor requests the resource' 
   it('should be denied')

describe('when a host account requests the resource'
    it('should receive the resource')
```

When written this way, it also reinforces the requirements that are being tested.

Meadowlark-js/tests/e2e/scenarios/SecurityValidation.test.ts
```diff
@@ -0,0 +1,372 @@
...
+
+    expect(contentClassDescriptorLocation).toBeTruthy();
+  });
+
+  afterAll(async () => {
+    await deleteResourceByLocation(contentClassDescriptorLocation);
+  });
+
+  // GET
+  describe('given that Vendor 1 creates an educationContent', () => {
```
stephenfuqua, line 98: We are using `EducationContent` here, but that is not the point of the test. Similarly, the test uses "vendor 1" but there is nothing special about that vendor, or about the `EducationContent` resource type. Therefore I recommend simplifying this description:
 ```suggestion
  describe('given a vendor API client has created a resource', () => {
```

**Review by stephenfuqua — COMMENTED** [review]
You have good tests here for the functionality. All of my comments are around structuring the test and describing what they are trying to accomplish - with the aims of 

* making it easier for a reader of the test to understand the business requirements covered by these scenarios, and
* reducing some redundancy in the CRUD scenarios to lower the test execution time

File: Meadowlark-js/tests/e2e/scenarios/ResourcesDeleteValidation.test.ts [hunk 4]
```diff
@@ -0,0 +1,69 @@
...
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+import { getAccessToken } from '../helpers/Credentials';
+import { createContentClassDescriptor } from '../helpers/DataCreation';
+import { deleteResourceByLocation } from '../helpers/Resources';
+import { baseURLRequest, rootURLRequest } from '../helpers/Shared';
+
+describe('given a DELETE of a non-existent EducationContent', () => {
```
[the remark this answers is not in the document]
DavidJGapCR: Thanks for the tip!

Meadowlark-js/tests/e2e/scenarios/ResourcesGetValidation.test.ts
DavidJGapCR: Just to confirm. I can create just one scenarios to cover all the CRUD methods. And I can use the Descriptor, no need to use the `EducationContent` resource. 
Is this correct?

stephenfuqua (approved):

PR #202, opened 2023-02-15
[RND-442] Resource Id as UUID for All Resources (MongoDB)

jleiva-gap:
- Replace documentId by MeadowlarkId.
- Add documentUuid as id.
- Add id to the body.
- Validate id with path (update).
- Add documentUuid to Update and Upsert files.
- Update code to filter by documentUuid.
- Add allowIdentityUpdates. If it's true, it deletes the old document and it inserts a new version to replace that document. If false, it tries to update the document if it has the same documentUuid and meadowlarkid.
- Update unit, integration and e2e test to match the new id.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My change requires a change to the documentation.
- [x] I have added tests to cover my changes.


Meadowlark-js/packages/meadowlark-core/src/handler/DescriptorLoader.ts
```diff
@@ -126,7 +126,8 @@ async function loadParsedDescriptors(descriptorData: XmlDescriptorData): Promise
       const documentInfo: DocumentInfo = descriptorDocumentInfoFrom(descriptorDocument);
 
       const upsertRequest: UpsertRequest = {
-        id: documentIdForDocumentInfo(resourceInfo, documentInfo),
+        id: getDocumentUuidForDocument(),
+        meadowlarkIdentity: documentIdForDocumentInfo(resourceInfo, documentInfo),
```
bradbanister, line 130: Let's call this `meadowlarkId` instead of `meadowlarkIdentity`. We use "Identity" to mean the un-hashed elements of a document, and "id" to mean the hashed version.

Meadowlark-js/packages/meadowlark-core/src/model/DocumentIdentity.ts
```diff
@@ -5,6 +5,7 @@
 
 import { normalizeDescriptorSuffix } from '@edfi/metaed-core';
 import crypto from 'node:crypto';
+import { v4 as uuidv4 } from 'uuid';
```
bradbanister, line 8: You don't need this. node:crypto has a UUID generator. https://nodejs.org/docs/latest-v16.x/api/crypto.html#cryptorandomuuidoptions

Meadowlark-js/backends/meadowlark-opensearch-backend/test/integration/UpdateOpensearch.test.ts
```diff
@@ -32,12 +32,15 @@ const resourceInfo: ResourceInfo = {
...
   isDescriptor: false,
   resourceVersion: '3.3.1-b',
+  allowIdentityUpdates: false,
 };
 
-const id = '1234a-5678b';
-
+const meadowlarkId = '1234a-5678b';
+// const documentUuid = getDocumentUuidForDocument();
+// TEMPORAL: Using meadowlarkId as documentUuid
```
andonyns, line 40: Out of curiosity, why is this temporal?

jleiva-gap: Because we are creating a documentUuid just for mongodb, but postgres still uses the meadowlarkId. As we use common types and to maintain compatibility, at this time in postgres the meadowlarkId and documentUuid would be taken as the same value.

The point is that it cannot be changed in this ticket because this has implications for keys and queries.

stephenfuqua: We'll need to change postgresql to use documentUuid as well at some point, right? Can you create a new ticket for that @jleiva-gap and put fix version {{meadowlark-0.4.0}}.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -50,7 +50,7 @@ export async function upsertDocument(
             blockingDocuments: [
               {
                 // eslint-disable-next-line no-underscore-dangle
-                documentId: superclassAliasIdInUse._id,
+                meadowlarkId: superclassAliasIdInUse._id,
```
stephenfuqua, line 53: When reporting on blocking documents, would it make sense to report on the `documentUuid` instead of the `meadowlarkId`, since it is the `documentUuid` that an API client would use to GET or DELETE that blocking item?

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/src/enhancer/JsonSchemaEnhancer.ts
```diff
@@ -42,6 +42,10 @@ const descriptorSchema: SchemaRoot = {
   title: 'EdFi.Descriptor',
   description: 'An Ed-Fi Descriptor',
   properties: {
+    id: {
+      type: 'string',
+      description: 'The item documentUuid.',
+    },
```
bradbanister, line 48: You should remove these. We don't include them because, per the [ODS/API documentation](https://api.ed-fi.org/v6.1/docs/swagger/index.html?urls.primaryName=Descriptors), including `id` in a body throws an error.

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/src/model/JsonSchema.ts
```diff
@@ -68,7 +68,12 @@ export function newSchemaRoot(): SchemaRoot {
     title: '',
     description: '',
     type: 'object',
-    properties: {},
+    properties: {
+      id: {
+        type: 'string',
+        description: 'The item id',
+      },
+    },
```
bradbanister, line 76: As above, should be removed

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/test/enhancer/JsonSchemaEnhancer.test.ts
```diff
@@ -2267,6 +2267,10 @@ describe('when building a descriptor', () => {
             "description": "The descriptor description",
             "type": "string",
           },
+          "id": {
+            "description": "The item documentUuid.",
+            "type": "string",
+          },
```
bradbanister, line 2273: As above, should be removed

Meadowlark-js/tests/e2e/scenarios/SecurityValidation.test.ts
```diff
@@ -220,6 +221,7 @@ describe('given the existance of two vendors and one host', () => {
 
     describe('when the same vendor updates the resource', () => {
       it('should return success', async () => {
+        resourceBodyPUTUpdated.id = resourceLocation.split('/').pop();
```
bradbanister, line 224: Would you mind renaming such variables to always be camel cased? In this case `resourceBodyPUTUpdated` -> `resourceBodyPutUpdated`

Meadowlark-js/tests/e2e/endpoints/educationContents.test.ts
```diff
@@ -64,10 +64,12 @@ describe('Education contents', () => {
 
     it('should edit an education content', async () => {
       const uri = 'uri://ed-fi.org/fake-updated-uri';
+      const id = educationContentLocation.split('/').pop();
       await rootURLRequest()
         .put(educationContentLocation)
         .auth(await getAccessToken('vendor'), { type: 'bearer' })
         .send({
+          id,
```
bradbanister, line 72: I think these test changes will need rolling back as well

jleiva-gap: @bradbanister , I need the id to compare the body id with the path component. As Stephen mentioned, ODS/API receives the surrogate key, in this case, our surrogate key = documentUuid. Then, if I don't send this value, PUT will return an Error.

bradbanister: I'll push a commit in a minute.

The old validation in update I think you are referring to is meaningless now. Previously, it validated that the identity of the document did not change even if other fields did, because we didn't allow that. Now we allow it, but also can't test for that in meadowlark-core anymore. Unfortunately, backends have to do it now via a lookup of meadowlarkId for a given documentUuid.

jleiva-gap: It makes sense. In some point we were replacing the value by documentUuid, so we could update a document using a complete different body, but if allowIdentityUpdates is false, we are filtering the replace with documentUuid and meadowlarkId.

Meadowlark-js/packages/meadowlark-core/src/model/DocumentIdentity.ts
```diff
@@ -74,10 +74,10 @@ function documentIdentityHashFrom(documentIdentity: DocumentIdentity): string {
  * The resulting Base64Url string is 38 characters long. The first 16 characters are the resource info hash and
  * the remaining 22 characters are the identity hash.
  */
-export function documentIdForDocumentIdentity(resourceInfo: BaseResourceInfo, documentIdentity: DocumentIdentity): string {
+export function meadowlarkIdForDocumentIdentity(resourceInfo: BaseResourceInfo, documentIdentity: DocumentIdentity): string {
   return `${resourceInfoHashFrom(resourceInfo)}${documentIdentityHashFrom(documentIdentity)}`;
 }
 
-export function getDocumentUuid(): string {
+export function generateDocumentUuid(): string {
```
bradbanister, line 81: Renamed because this function makes something new.

Meadowlark-js/packages/meadowlark-core/src/model/DocumentIdentity.ts
```diff
@@ -74,10 +74,10 @@ function documentIdentityHashFrom(documentIdentity: DocumentIdentity): string {
  * The resulting Base64Url string is 38 characters long. The first 16 characters are the resource info hash and
  * the remaining 22 characters are the identity hash.
  */
-export function documentIdForDocumentIdentity(resourceInfo: BaseResourceInfo, documentIdentity: DocumentIdentity): string {
+export function meadowlarkIdForDocumentIdentity(resourceInfo: BaseResourceInfo, documentIdentity: DocumentIdentity): string {
   return `${resourceInfoHashFrom(resourceInfo)}${documentIdentityHashFrom(documentIdentity)}`;
```
bradbanister, line 78: Renamed for clarity that this returns a meadowlarkId.

Meadowlark-js/packages/meadowlark-core/src/model/DocumentInfo.ts
```diff
@@ -54,23 +54,8 @@ export const NoDocumentInfo = Object.freeze({
 });
 
 /**
- * Returns the id, using the uuid hash.
+ * Returns the meadowlarkId of the given DocumentInfo, using the project name, resource name and identity of the API document.
  */
-export function getDocumentUuidForDocument(): string {
-  return getDocumentUuid();
-}
```
bradbanister: No longer of value

Meadowlark-js/packages/meadowlark-core/src/model/DocumentInfo.ts
```diff
@@ -54,23 +54,8 @@ export const NoDocumentInfo = Object.freeze({
...
-export function documentIdForDocumentInfo(resourceInfo: ResourceInfo, documentInfo: DocumentInfo): string {
-  return documentIdForDocumentIdentity(resourceInfo, documentInfo.documentIdentity);
-}
-
-/**
- * Returns the id from the body.
- */
-export function documentUuidForDocumentBody(parsedBody: object): string {
-  const jsonFromBody = JSON.parse(JSON.stringify(parsedBody));
-  return jsonFromBody.id;
```
bradbanister: Dropped id from body

Meadowlark-js/packages/meadowlark-core/src/model/DocumentInfo.ts
```diff
@@ -54,23 +54,8 @@ export const NoDocumentInfo = Object.freeze({
...
+ * Returns the meadowlarkId of the given DocumentInfo, using the project name, resource name and identity of the API document.
  */
-export function getDocumentUuidForDocument(): string {
-  return getDocumentUuid();
-}
-
-/**
- * Returns the id of the given DocumentInfo, using the project name, resource name and identity of the API document.
- */
-export function documentIdForDocumentInfo(resourceInfo: ResourceInfo, documentInfo: DocumentInfo): string {
```
bradbanister, line 59: Renamed for clarity that this returns a meadowlarkId

Meadowlark-js/packages/meadowlark-core/test/handler/Update.test.ts
```diff
@@ -137,51 +137,3 @@ describe('given the update succeeds', () => {
     expect(response.body).toBeUndefined();
   });
 });
-
-describe('given the resourceId of the update does not match the id derived from the body', () => {
-  let mockDocumentStore: any;
```
bradbanister, line 142: meadowlark-core can no longer do this

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/src/enhancer/JsonSchemaEnhancer.ts
```diff
@@ -42,10 +42,6 @@ const descriptorSchema: SchemaRoot = {
   title: 'EdFi.Descriptor',
   description: 'An Ed-Fi Descriptor',
   properties: {
-    id: {
-      type: 'string',
-      description: 'The item documentUuid.',
-    },
```
bradbanister: dropping id, shouldn't be valid

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/src/model/JsonSchema.ts
```diff
@@ -68,12 +68,7 @@ export function newSchemaRoot(): SchemaRoot {
     title: '',
     description: '',
     type: 'object',
-    properties: {
-      id: {
-        type: 'string',
-        description: 'The item id',
-      },
-    },
+    properties: {},
```
bradbanister, line 71: dropping id, shouldn't be valid

Meadowlark-js/packages/metaed-plugin-edfi-meadowlark/test/enhancer/JsonSchemaEnhancer.test.ts
```diff
@@ -2267,10 +2267,6 @@ describe('when building a descriptor', () => {
             "description": "The descriptor description",
             "type": "string",
           },
-          "id": {
-            "description": "The item documentUuid.",
-            "type": "string",
-          },
```
bradbanister: dropping id, shouldn't be valid

Meadowlark-js/packages/meadowlark-core/src/middleware/ValidateResourceIdMiddleware.ts
```diff
@@ -21,27 +20,7 @@ export async function resourceIdValidation({
   writeRequestToLog(moduleName, frontendRequest, 'resourceIdValidation');
 
   // If no resource id in request, there is nothing to validate
-  if (frontendRequest.middleware.pathComponents.resourceId == null) return { frontendRequest, frontendResponse: null };
-
```
bradbanister, line 25: This whole middleware and its test can be deleted. It doesn't do anything now.

Meadowlark-js/packages/meadowlark-core/src/handler/Delete.ts
```diff
@@ -27,13 +28,17 @@ export async function deleteIt(frontendRequest: FrontendRequest): Promise<Fronte
...
     }
-
+    const meadowlarkId = meadowlarkIdForDocumentInfo(
+      frontendRequest.middleware.resourceInfo,
+      frontendRequest.middleware.documentInfo,
+    );
     const request: DeleteRequest = {
-      id: frontendRequest.middleware.pathComponents.resourceId,
+      documentUuid: frontendRequest.middleware.pathComponents.documentUuid,
+      meadowlarkId,
```
bradbanister, line 41: There is no body for a delete request, so there is no document to extract a DocumentIdentity from, so the meadowlarkId that can be computed is useless (only has the resource information). It should be removed from DeleteRequest.

Meadowlark-js/packages/meadowlark-core/src/message/DeleteRequest.ts
```diff
@@ -7,7 +7,8 @@ import { ResourceInfo } from '../model/ResourceInfo';
 import { Security } from '../security/Security';
 
 export type DeleteRequest = {
-  id: string;
+  documentUuid: string;
+  meadowlarkId: string;
```
bradbanister, line 11: There is no body for a delete request, so there is no document to extract a DocumentIdentity from, so the meadowlarkId that can be computed is useless (only has the resource information). It should be removed from DeleteRequest.

Meadowlark-js/packages/meadowlark-core/src/message/GetRequest.ts
```diff
@@ -7,7 +7,8 @@ import { ResourceInfo } from '../model/ResourceInfo';
 import { Security } from '../security/Security';
 
 export type GetRequest = {
-  id: string;
+  documentUuid: string;
+  meadowlarkId: string;
```
bradbanister, line 11: There is no body for a get request, so there is no document to extract a DocumentIdentity from, so the meadowlarkId that can be computed is useless (only has the resource information). It should be removed from GetRequest.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -79,15 +80,44 @@ export async function updateDocumentById(
...
-      Logger.debug(`${moduleName}.updateDocumentById: Updating document id ${id}`, traceId);
+      Logger.debug(`${moduleName}.updateDocumentById: Updating document uuid ${documentUuid}`, traceId);
 
-      const { acknowledged, matchedCount } = await mongoCollection.replaceOne({ _id: id }, document, { session });
-      if (acknowledged) {
-        updateResult.response = matchedCount > 0 ? 'UPDATE_SUCCESS' : 'UPDATE_FAILURE_NOT_EXISTS';
+      // if resourceInfo.allowIdentityUpdates is true, the process deletes de old document and inserts a new document.
+      if (resourceInfo.allowIdentityUpdates) {
+        const deleteResult = await deleteDocumentById(
+          { documentUuid, meadowlarkId, resourceInfo, security, validate: true, traceId },
```
bradbanister, line 88: This function is at too high a level to use here. It manages its own transaction, so the `delete` transaction will be outside the `update` transaction.

jleiva-gap: If in this option I delete a document, but I can't insert the new version, since the upsert and delete are in different transactions, should I reinsert the original document?

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/OwnershipSecurity.ts
```diff
@@ -36,35 +39,44 @@ export async function rejectByOwnershipSecurity(
...
+  let result: WithId<MeadowlarkDocument> | null = null;
 
   try {
-    // if documentUuid is null, it uses _id
-    const result: WithId<MeadowlarkDocument> | null =
-      documentUuid != null
-        ? await mongoCollection.findOne({ documentUuid }, { projection: { createdBy: 1, _id: 0 } })
-        : await mongoCollection.findOne({ _id: meadowlarkId }, { projection: { createdBy: 1, _id: 0 } });
-    if (result === null) {
-      Logger.debug(`${functionName} - document not found for documentUuid ${documentUuid}`, frontendRequest.traceId);
```
bradbanister: Rearranged mostly for more accurate log messages

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -83,9 +83,13 @@ export async function updateDocumentById(
...
 
       // if resourceInfo.allowIdentityUpdates is true, the process deletes de old document and inserts a new document.
+
+      // TODO ******
+      // /////////////////////////////////////////  this needs to be in the same transaction
+      //
       if (resourceInfo.allowIdentityUpdates) {
         const deleteResult = await deleteDocumentById(
-          { documentUuid, meadowlarkId, resourceInfo, security, validate: true, traceId },
+          { documentUuid, resourceInfo, security, validate: true, traceId },
```
bradbanister, line 92: This need to be direct MongoDB call(s) that use the transaction Session object

Meadowlark-js/backends/meadowlark-mongodb-backend/test/integration/Delete.test.ts
```diff
@@ -17,36 +17,40 @@ import {
...
-import crypto from 'node:crypto';
 import { MeadowlarkDocument } from '../../src/model/MeadowlarkDocument';
 import { getDocumentCollection, getNewClient } from '../../src/repository/Db';
 import { deleteDocumentById } from '../../src/repository/Delete';
 import { upsertDocument } from '../../src/repository/Upsert';
 import { setupConfigForIntegration } from './Config';
 
 jest.setTimeout(40000);
-const documentUuid = '3518d452-a7b7-4f1c-aa91-26ccc48cf4b8';
-const documentUuidBackup = '4518d452-a7b7-4f1c-aa91-26ccc48cf4b8';
```
bradbanister: "backup" doesn't seem like an accurate description. It's really just a 2nd id.

Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/OpenSearchException.ts
```diff
@@ -6,7 +6,7 @@ import { OpenSearchClientError, ResponseError } from '@opensearch-project/opense
 import { Logger } from '@edfi/meadowlark-utilities';
 import { QueryResult } from '@edfi/meadowlark-core';
 
-export async function LogOpenSearchErrors(
+export async function handleOpenSearchError(
```
bradbanister, line 9: renamed because it doesn't only log. it also decides on a result to return in the case of an error.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Get.ts
```diff
@@ -4,13 +4,15 @@
...
+import { GetResult, GetRequest, MeadowlarkId } from '@edfi/meadowlark-core';
 import { Logger } from '@edfi/meadowlark-utilities';
 import { findDocumentByIdSql } from './SqlHelper';
 
 const moduleName = 'postgresql.repository.Get';
 
-export async function getDocumentById({ meadowlarkId, traceId }: GetRequest, client: PoolClient): Promise<GetResult> {
+export async function getDocumentById({ documentUuid, traceId }: GetRequest, client: PoolClient): Promise<GetResult> {
+  // TODO *** cast to unknown is an invalid mixing of meadowlarkId and documentUuid
+  const meadowlarkId: MeadowlarkId = documentUuid as unknown as MeadowlarkId;
```
bradbanister, line 15: This is wrong, but we aren't worrying about PostgreSQL right now. The nice thing is that a future search for "as unknown as" will find all the problem areas.

Meadowlark-js/package.json
```diff
@@ -46,11 +46,11 @@
     "test:lint": "npm run test:lint:ts && npm run test:lint:eslint",
     "test:lint:eslint": "eslint --max-warnings 0 --ext .js,.ts .",
     "test:lint:ts": "tsc -p . --noEmit",
-    "test:unit": "cross-env NODE_OPTIONS=--max-old-space-size=4096 LOG_LEVEL=warn jest --projects tests/config/unit",
-    "test:integration": "cross-env NODE_OPTIONS=--max-old-space-size=4096 LOG_LEVEL=warn jest --projects tests/config/integration --runInBand",
+    "test:unit": "cross-env NODE_OPTIONS=--max-old-space-size=6144 LOG_LEVEL=warn jest --projects tests/config/unit",
+    "test:integration": "cross-env NODE_OPTIONS=--max-old-space-size=6144 LOG_LEVEL=warn jest --projects tests/config/integration --runInBand",
```
bradbanister, line 50: Tried upping max memory for tests to 6MB instead of 4MB, now that we have worker killer in place.

Meadowlark-js/packages/meadowlark-core/src/handler/FrontendFacade.ts
```diff
@@ -64,7 +63,6 @@ function putStack(): MiddlewareStack {
       R.andThen(parseBody),
       R.andThen(anonymizeAndLogRequestBody),
       R.andThen(resourceValidation),
-      R.andThen(resourceIdValidation),
```
bradbanister, line 67: No longer does anything

Meadowlark-js/packages/meadowlark-core/src/handler/GetById.ts
```diff
@@ -26,13 +26,9 @@ export async function getById(frontendRequest: FrontendRequest): Promise<Fronten
   }
   const request: GetRequest = {
     documentUuid: frontendRequest.middleware.pathComponents.documentUuid,
-    meadowlarkId: meadowlarkIdForDocumentInfo(
-      frontendRequest.middleware.resourceInfo,
-      frontendRequest.middleware.documentInfo,
-    ),
```
bradbanister: No body to get a meadowlarkId from

Meadowlark-js/packages/meadowlark-core/src/message/UpdateRequest.ts
```diff
@@ -3,17 +3,6 @@
...
-  meadowlarkId: string;
-  documentUuid: string;
-  resourceInfo: ResourceInfo;
-  documentInfo: DocumentInfo;
-  edfiDoc: object;
-  validate: boolean;
-  security: Security;
-  traceId: string;
-};
+export type UpdateRequest = UpsertRequest;
```
bradbanister, line 8: They are the same shape, and have been for a long time. Keeping separate names in case they become different in the future.

Meadowlark-js/packages/meadowlark-core/src/model/DocumentInfo.ts
```diff
@@ -52,10 +51,3 @@ export function newDocumentInfo(): DocumentInfo {
 export const NoDocumentInfo = Object.freeze({
   ...newDocumentInfo(),
 });
-
-/**
- * Returns the meadowlarkId of the given DocumentInfo, using the project name, resource name and identity of the API document.
- */
-export function meadowlarkIdForDocumentInfo(resourceInfo: ResourceInfo, documentInfo: DocumentInfo): string {
-  return meadowlarkIdForDocumentIdentity(resourceInfo, documentInfo.documentIdentity);
-}
```
bradbanister, line 61: This function no longer does enough to be worth having it.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/OwnershipSecurity.ts
```diff
@@ -35,30 +38,45 @@ export async function rejectByOwnershipSecurity(
...
 
   try {
-    const result: WithId<MeadowlarkDocument> | null = await mongoCollection.findOne(
-      { _id: id },
-      { projection: { createdBy: 1, _id: 0 } },
-    );
-    if (result === null) {
-      Logger.debug(`${functionName} - document not found for id ${id}`, frontendRequest.traceId);
-      return 'NOT_APPLICABLE';
+    if (documentUuid != null) {
```
stephenfuqua, line 46: Something is not working correctly in the ownership security.

Try manually testing with file `local.33b.http`.  Create an `EducationContents` records (line 119). Try the GET ALL that follows it. In my case, this _does_ include the new record. Note that it is an OpenSearch query. Then try the GET BY ID that follows that (line 136). I saw it succeed once, and then fail repeatedly. The PUT request also failed. Interestingly,  rerunning the POST request works.

```
2023-02-28 08:21:30.802 info 1564839f-36f2-474f-be08-87e80939f564 mongodb.repository.OwnershipSecurity.rejectByOwnershipSecurity /v3.3b/ed-fi/EducationContents/1c6674ba-64d2-4068-beb9-b21fb4ff124a
2023-02-28 08:21:30.804 debug 1564839f-36f2-474f-be08-87e80939f564 mongodb.repository.OwnershipSecurity.rejectByOwnershipSecurity - document not found for documentUuid 1c6674ba-64d2-4068-beb9-b21fb4ff124a
```

Meadowlark-js/packages/meadowlark-core/src/model/DocumentIdentity.ts
```diff
@@ -74,6 +75,13 @@ function documentIdentityHashFrom(documentIdentity: DocumentIdentity): string {
...
-  return `${resourceInfoHashFrom(resourceInfo)}${documentIdentityHashFrom(documentIdentity)}`;
+export function meadowlarkIdForDocumentIdentity(
+  resourceInfo: BaseResourceInfo,
+  documentIdentity: DocumentIdentity,
+): MeadowlarkId {
+  return `${resourceInfoHashFrom(resourceInfo)}${documentIdentityHashFrom(documentIdentity)}` as MeadowlarkId;
+}
+
+export function generateDocumentUuid(): DocumentUuid {
+  return crypto.randomUUID() as DocumentUuid;
```
stephenfuqua, line 86: If you run the same POST request twice, then each one is generating a new resource with a different `DocumentUuid`, even though the rest of the stored document is the same.

When doing an UPSERT on an previously stored document, we _must_ keep the existing identifier instead of assigning a new one.

stephenfuqua (changes requested):

Meadowlark-js/packages/meadowlark-core/src/model/DocumentIdentity.ts
jleiva-gap: We found a problem with the documentUuid, the process created a new uuid each time, but that uuid didn't have a relationship with a document. The code was updated to fix this.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/OwnershipSecurity.ts
jleiva-gap: This problem is related to the POST returning different uuids each time. Now it is fixed

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -8,80 +8,117 @@ import { Logger } from '@edfi/meadowlark-utilities';
...
-
-      const { acknowledged, matchedCount } = await mongoCollection.replaceOne({ _id: id }, document, { session });
+      Logger.debug(`${moduleName}.updateDocumentById: Updating document uuid ${documentUuid}`, traceId);
+      const { acknowledged, matchedCount } = await mongoCollection.replaceOne(
+        { _id: meadowlarkId, documentUuid },
+        document,
+        {
+          session,
+        },
+      );
```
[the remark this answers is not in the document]
bradbanister: I think I'd prefer not to have the getById happen in the handler because some backends might be able to optimize the check themselves and avoid a second query.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -8,80 +8,117 @@ import { Logger } from '@edfi/meadowlark-utilities';
...
+      security.clientId,
+    );
+    // if resourceInfo.allowIdentityUpdates is true, the process deletes de old document and inserts a new document.
+    if (resourceInfo.allowIdentityUpdates) {
+      const deleteResult = await deleteDocumentByIdTransaction(
+        { documentUuid, resourceInfo, security, validate: true, traceId },
+        client,
+        session,
+      );
+      // if the document was deleted, it should insert the new version.
```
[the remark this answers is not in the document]
bradbanister: ➕

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -8,80 +8,117 @@ import { Logger } from '@edfi/meadowlark-utilities';
...
+      if (deleteResult.response === 'DELETE_SUCCESS') {
+        // insert the updated document.
+        const upsertResult = await upsertDocumentTransaction(
```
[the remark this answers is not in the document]
bradbanister: This looks half-right. The `session` object is being passed down here (line 80 and 97), but the `session.withTransaction()` shouldn't be in those functions. It needs to be outside those functions if the transaction is to be shared, like it was around line 30 before this.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -8,80 +8,117 @@ import { Logger } from '@edfi/meadowlark-utilities';
...
-        edfiDoc,
-        validate,
-        security.clientId,
-      );
-
+      return updateResult;
+    }
+    // Start update transaction
+    await session.withTransaction(async () => {
       await writeLockReferencedDocuments(mongoCollection, document.outboundRefs, session);
```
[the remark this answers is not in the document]
bradbanister: ➕

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Delete.ts
```diff
@@ -87,11 +85,13 @@ export async function deleteDocumentById(
...
   let deleteResult: DeleteResult = { response: 'UNKNOWN_FAILURE', failureMessage: '' };
   try {
-    deleteResult = await deleteDocumentByIdTransaction(
-      { documentUuid, validate, traceId } as DeleteRequest,
-      client,
-      session,
-    );
+    await session.withTransaction(async () => {
+      deleteResult = await deleteDocumentByIdTransaction(
+        { documentUuid, validate, traceId } as DeleteRequest,
```
bradbanister, line 90: This isn't a full DeleteRequest, and shouldn't be cast to one.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Delete.ts
```diff
@@ -20,63 +20,61 @@ export async function deleteDocumentByIdTransaction(
 ): Promise<DeleteResult> {
   const mongoCollection: Collection<MeadowlarkDocument> = getDocumentCollection(client);
```
bradbanister, line 21: This function only ever uses `client` to get the `mongoCollection`, so better to just pass in a `Collection<MeadowlarkDocument>`

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
-      updateResult = { response: 'UPDATE_FAILURE_NOT_EXISTS' };
-    }
-  } else if (deleteResult.response === 'DELETE_FAILURE_NOT_EXISTS') {
-    updateResult = { response: 'UPDATE_FAILURE_NOT_EXISTS' };
+        'Got "UPDATE_SUCCESS" from upsertDocumentTransaction() but document should have been deleted first',
+      );
+      return { response: 'UPDATE_SUCCESS' };
+    case 'INSERT_FAILURE_REFERENCE':
+      // Something unexpected happened. Validation should not have occurred.
+      Logger.error(
```
bradbanister, line 60: Needed to handle all possible cases

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
-    } else {
-      const msg =
-        'mongoCollection.replaceOne returned acknowledged: false, indicating a problem with write concern configuration';
-      Logger.error(`${moduleName}.updateDocumentById`, traceId, msg);
-    }
+  const { documentUuid, resourceInfo, traceId, security } = updateRequest;
+
+  // Optimize by trying a replacement update, which will succeed if there is no identity change
+  const tryUpdateByReplacementResult: UpdateResult | null = await tryUpdateByReplacement(
+    document,
```
bradbanister, line 130: First db call attempts the update as normal, figuring a non-identity change is the common case. Only If it's an identity change or not-exist error do we pay the second db call penalty.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
+    updateRequest,
+    mongoCollection,
+    session,
+  );
+
+  if (tryUpdateByReplacementResult != null) {
+    // Ensure referenced documents are not modified in other transactions
+    await writeLockReferencedDocuments(mongoCollection, document.outboundRefs, session);
+    return tryUpdateByReplacementResult;
```
bradbanister, line 139: Only write-lock if the attempt succeeds. Otherwise, let upsert do the write locking.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
+  }
+
+  // Either the documentUuid doesn't exist or the identity has changed.
+  // The following delete attempt will catch if documentUuid does not exist
+  const deleteResult = await deleteDocumentByIdTransaction(
+    { documentUuid, resourceInfo, security, validateNoReferencesToDocument: true, traceId },
+    mongoCollection,
+    session,
+  );
```
bradbanister, line 148: We can catch the non-exist condition with the delete db call.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
+      return { response: 'UPDATE_FAILURE_NOT_EXISTS' };
+    case 'DELETE_FAILURE_REFERENCE':
+      // We have an update cascade scenario
+      //
+      // TODO: Collect all the blocking documents that need updating (not just the first 5!)
+      //       Delete again without validating references
+      //       Do the insert
+      //       If everything succeeds, schedule the cascade - updating each referring document
+      // For now, let it be known this is an update cascade scenario, but do nothing
+      return { response: 'UPDATE_CASCADE_REQUIRED' };
```
bradbanister, line 166: This is the update cascade condition we'll support in a future ticket.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
+  // Ensure referenced documents are not modified in other transactions
+  await writeLockReferencedDocuments(mongoCollection, document.outboundRefs, session);
+
+  const tryUpdateByReplacementResult: UpdateResult | null = await tryUpdateByReplacement(
+    document,
+    updateRequest,
+    mongoCollection,
+    session,
+  );
+
```
bradbanister, line 199: Again, try the common case first. If it doesn't work, then make a second call to find out why.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
+  const documentByUuid: WithId<MeadowlarkDocument> | null = await mongoCollection.findOne(
+    { documentUuid: updateRequest.documentUuid },
+    onlyReturnId(session),
+  );
+
+  if (documentByUuid == null) return { response: 'UPDATE_FAILURE_NOT_EXISTS' };
+
+  // The identity of the new document is different from the existing document
+  return { response: 'UPDATE_FAILURE_IMMUTABLE_IDENTITY' };
+}
```
bradbanister, line 213: New response type when identity can't be changed.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -3,96 +3,273 @@
...
+
+/**
+ * Checks for any invalid references in document. If found, provide an UpdateResult with information on the errors.
+ *
+ * This function expects Session to have an active transaction. Aborting the transaction on error is left to the caller.
+ */
+async function checkForInvalidReferences(
+  { documentInfo, documentUuid, traceId, meadowlarkId }: UpdateRequest,
```
bradbanister, line 221: extracted validation to make updateDocumentById() smaller

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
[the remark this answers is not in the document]
bradbanister: It ended up that ordering can be optimized if it stays in the backend.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -3,64 +3,82 @@
...
+    documentUuidForInsert,
+    meadowlarkId,
+    edfiDoc,
+    validateDocumentReferencesExist,
+    traceId,
+    security,
+  }: UpsertRequest,
+  mongoCollection: Collection<MeadowlarkDocument>,
   session: ClientSession,
+  documentFromUpdate?: MeadowlarkDocument,
```
bradbanister, line 36: If this call is coming from updateByDocumentId, it's already created the MeadowlarkDocument. This avoids a recalc of the meadowlarkIds of document references.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -3,64 +3,82 @@
...
-
 const moduleName: string = 'mongodb.repository.Upsert';
+
 export async function upsertDocumentTransaction(
-  { resourceInfo, documentInfo, documentUuidInserted, meadowlarkId, edfiDoc, validate, traceId, security }: UpsertRequest,
-  client: MongoClient,
+  {
+    resourceInfo,
+    documentInfo,
+    documentUuidForInsert,
```
bradbanister, line 27: There are two different documentUuids at play here. This one is provided for an insert situation, either speculatively by the meadowlark-core handler, or by updateByDocumentId when it's an insert for identity change reasons.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -3,64 +3,82 @@
...
-  let upsertResult: UpsertResult = { response: 'UNKNOWN_FAILURE' };
-  let documentUuid = documentUuidInserted;
+  // Check whether this document exists in the db
+  const existingDocument: WithId<MeadowlarkDocument> | null = await mongoCollection.findOne(
+    { _id: meadowlarkId },
+    onlyReturnDocumentUuid(session),
+  );
+
+  // the documentUuid of the existing document if this is an update, or null if this is an insert
+  const existingDocumentUuid: DocumentUuid | null = existingDocument?.documentUuid ?? null;
```
bradbanister, line 45: This is the other documentUuid at play, the (optional) one where this is an update condition.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -3,64 +3,82 @@
...
   if (isInsert && documentInfo.superclassInfo != null) {
     const superclassAliasId: string = documentIdForSuperclassInfo(documentInfo.superclassInfo);
-    const superclassAliasIdInUse: WithId<MeadowlarkDocument> | null = await mongoCollection.findOne({
-      aliasIds: superclassAliasId,
-    });
+    const superclassAliasIdInUse: WithId<MeadowlarkDocument> | null = await mongoCollection.findOne(
+      {
+        aliasIds: superclassAliasId,
+      },
+      { session },
```
bradbanister, line 57: Not sure why this wasn't in the session

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
bradbanister: Note there is no reason to make this db call if this call is coming from updateByDocumentId. I ran out of time, and a minor optimization since it's probably not the common case.

Meadowlark-js/packages/meadowlark-core/src/handler/Update.ts
```diff
@@ -65,13 +65,27 @@ export async function update(frontendRequest: FrontendRequest): Promise<Frontend
...
-        body: R.is(String, failureMessage) ? { error: failureMessage, blockingUris } : failureMessage,
+        body: R.is(String, result.failureMessage) ? { error: result.failureMessage, blockingUris } : result.failureMessage,
         statusCode: 409,
         headers: headerMetadata,
       };
     }
 
-    writeErrorToLog(moduleName, frontendRequest.traceId, 'update', 500, failureMessage);
+    if (response === 'UPDATE_CASCADE_REQUIRED') {
+      writeDebugStatusToLog(moduleName, frontendRequest, 'update', 409, 'update cascade required');
```
bradbanister, line 75: Handle the new cases

Meadowlark-js/packages/meadowlark-core/src/message/DeleteRequest.ts
```diff
@@ -10,7 +10,7 @@ import { Security } from '../security/Security';
 export type DeleteRequest = {
   documentUuid: DocumentUuid;
   resourceInfo: ResourceInfo;
-  validate: boolean;
+  validateNoReferencesToDocument: boolean;
   security: Security;
```
bradbanister, line 14: I realized while working on update that our `validate` term covers two different kinds of validations. This one is that nothing references the document to be deleted.

Meadowlark-js/packages/meadowlark-core/src/message/UpdateRequest.ts
```diff
@@ -14,7 +14,7 @@ export type UpdateRequest = {
   resourceInfo: ResourceInfo;
   documentInfo: DocumentInfo;
   edfiDoc: object;
-  validate: boolean;
+  validateDocumentReferencesExist: boolean;
```
bradbanister, line 17: This one is that the references in the document body actually exist.

stephenfuqua (commented):
There is still a problem somewhere... if I post the same contents twice in a row, two resources are created, with different documentUuids.

stephenfuqua (approved):

PR #209, opened 2023-02-28
[RND-500] Fix e2e connection issues

andonyns:
These are a series of improvements for the e2e tests.

- Change ports to not crash with local env. This will allow to have the local environment next to the e2e tests.
- Create a single replica set to simplify setup process and not depend on temp container and shared volume.
- Refactor pipeline to avoid doing steps that will be done by the docker compose file.
- Wait for health check on open search container to avoid running if not ready.
- Adding additional log messages.


stephenfuqua (commented):
Code looks good to me, though I'll encourage someone else to test, approve, and merge it.

stephenfuqua (approved):

PR #212, opened 2023-03-02
[RND-401] - DRAFT - Adds retry logic for MongoDB.

DavidJGapCR:
## Types of changes

## Checklist

- [ ] I have signed all of the commits on this PR.
- [ ] I have agreed to the Ed-Fi Individual Contributors' License
- [ ] My code follows the code style of this project.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] All new and existing tests passed.


Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -113,9 +114,27 @@ export async function upsertDocument(
...
-      if (acknowledged) {
-        upsertResult.response = upsertedCount === 0 ? 'UPDATE_SUCCESS' : 'INSERT_SUCCESS';
+      const numberOfRetries: number = Config.get('MAX_NUMBER_OF_RETRIES') ?? 1;
+      const mongoUpsertResult = {
+        acknowledged: false,
+        upsertedCount: 0,
+      };
+
+      await retry(
+        async () => {
```
bradbanister, line 124: I'm thinking you should instead abort and retry the whole transaction when any mongo call in the transaction throws an exception.

Separately, you'll want to sync with main, as RND-442 made a lot of changes, one of which is to extract much of the transaction block of upsert and update.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -125,6 +144,13 @@ export async function upsertDocument(
   } catch (e) {
     Logger.error(`${moduleName}.upsertDocument`, traceId, e);
 
+    if (e.message === '[MongoServerError: WriteConflict]') {
```
bradbanister, line 147: This is good, but are there any other indications in the error object we can use, apart from a string compare? For example, what if the message changes slightly in a new driver version?

Can you check the type of `e`, maybe it's a `MongoServerError` and there are additional fields on it besides `message`. You may need to look at the TypeScript type definitions to see, or trigger the scenario and look at the exception in debug.

DavidJGapCR: Yes, that is a good point. Thanks.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
DavidJGapCR: Ok, sounds good.
Yes. I am working on the rebase. I'll probably end up closing this pull request and opening a new one.

PR #216, opened 2023-03-08
[RND-401] Adds retry logic for MongoDB

DavidJGapCR:


Meadowlark-js/packages/meadowlark-core/test/handler/Update.test.ts
```diff
@@ -108,6 +108,42 @@ describe('given the new document has an invalid reference ', () => {
...
+
+  it('has a failure message', () => {
+    expect(response.body).toMatchInlineSnapshot(`
+      {
+        "error": "Write conflict error returned",
+      }
+    `);
+  });
+});
+
```
bradbanister, line 146: As mentioned in the ticket, do we need a test for the case where 404 should be returned if a PUT fails due to a DELETE? Or is it already covered?

Meadowlark-js/packages/meadowlark-core/test/handler/DeleteIt.test.ts
```diff
@@ -52,6 +52,41 @@ describe('given there is no resourceId', () => {
...
+  let response: FrontendResponse;
+  let mockDocumentStore: any;
+
+  beforeAll(async () => {
+    mockDocumentStore = jest.spyOn(PluginLoader, 'getDocumentStore').mockReturnValue({
+      ...NoDocumentStorePlugin,
+      deleteDocumentById: async () =>
+        Promise.resolve({
+          response: 'DELETE_FAILURE_WRITE_CONFLICT',
+          failureMessage: 'Write conflict error returned',
```
bradbanister, line 65: While this message is certainly true, I think there could be a better API client-facing message for these 409s. There should be an explanation why. Something like "Write conflict due to concurrent access to this or related resources."

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -254,64 +255,87 @@ async function checkForInvalidReferences(
   };
 }
 
+async function updateDocumentByIdTransaction(
+  updateRequest: UpdateRequest,
```
bradbanister, line 259: I extracted the transaction behavior into a separate function that mirrors the structure of upsert and delete, to simplify the code wrapped in the retry block.

Meadowlark-js/packages/meadowlark-core/test/handler/Update.test.ts
DavidJGapCR: I don't think there is a way to know that a transaction is failing due to a delete. I mean, Mongo will throw the error when there is a conflict, but it won't tell me if the error occurred because the transaction being processed is a delete or an update. I think I can inspect the mongo error object to double check.

bradbanister:
One last thing, is this configured and/or default with some kind of backoff? We'll probably want that. Don't think that's something we need to expose as an env var though.

DavidJGapCR:
> One last thing, is this configured and/or default with some kind of backoff? We'll probably want that. Don't think that's something we need to expose as an env var though.

Sorry, I don't think I understand what you mean with **backoff**. 

The repositories will retry once, if the value is not configured.

Do you mean we should configure this value somewhere else, not the env file?

Meadowlark-js/.env.example
```diff
@@ -27,4 +27,7 @@ POSTGRES_PORT=5432
 
 # Do not override MongoDB settings here, because the integration tests use an "in memory"
 # version of MongoDB that is configured automatically. Any settings applied here will
-# cause the connection to fail.
\ No newline at end of file
+# cause the connection to fail.
```
stephenfuqua, line 30: The above message should be changed, since we're clearly overriding a MongoDB setting.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Update.ts
```diff
@@ -254,64 +255,87 @@ async function checkForInvalidReferences(
...
-        updateResult = await updateAllowingIdentityChange(document, updateRequest, mongoCollection, session);
-      } else {
-        updateResult = await updateDisallowingIdentityChange(document, updateRequest, mongoCollection, session);
-      }
-
-      if (updateResult.response !== 'UPDATE_SUCCESS') {
-        await session.abortTransaction();
-      }
-    });
+    const numberOfRetries: number = Config.get('MONGODB_MAX_NUMBER_OF_RETRIES') ?? 1;
```
stephenfuqua, line 306: This is why we want that default value to be set in the `Config` module - so that we don't need to repeat ourselves every time we access the config value.

Meadowlark-js/backends/meadowlark-mongodb-backend/src/repository/Upsert.ts
```diff
@@ -157,15 +158,39 @@ export async function upsertDocument(upsertRequest: UpsertRequest, client: Mongo
   const session: ClientSession = client.startSession();
   let upsertResult: UpsertResult = { response: 'UNKNOWN_FAILURE' };
   try {
-    await session.withTransaction(async () => {
-      upsertResult = await upsertDocumentTransaction(upsertRequest, mongoCollection, session);
-      if (upsertResult.response !== 'UPDATE_SUCCESS' && upsertResult.response !== 'INSERT_SUCCESS') {
-        await session.abortTransaction();
-      }
-    });
+    const numberOfRetries: number = Config.get('MONGODB_MAX_NUMBER_OF_RETRIES') ?? 1;
```
stephenfuqua, line 161: Another one

Meadowlark-js/packages/meadowlark-utilities/src/Config.ts
```diff
@@ -156,4 +157,5 @@ export async function initializeConfig(provider: ConfigPlugin) {
 
   set('BEGIN_ALLOWED_SCHOOL_YEAR', await provider.getInt('BEGIN_ALLOWED_SCHOOL_YEAR', 1900));
   set('END_ALLOWED_SCHOOL_YEAR', await provider.getInt('END_ALLOWED_SCHOOL_YEAR', 2100));
+  set('MONGODB_MAX_NUMBER_OF_RETRIES', await provider.getInt('MONGODB_MAX_NUMBER_OF_RETRIES', 1));
```
stephenfuqua, line 160: The default value of `1` is already being set correctly.

docs/CONFIGURATION.md
```diff
@@ -55,6 +55,7 @@ release 0.3.0.
 | -------------------------- | ---------- | -------------------------------------------------------------------------------------------------------------------------------------- |
 | MEADOWLARK_DATABASE_NAME   | meadowlark | For both MongoDB and PostgreSQL                                                                                                        |
 | **MONGO_URI**                  | (none)     | [Connection string URI](https://www.mongodb.com/docs/v4.0/reference/connection-string/) for accessing a MongoDB 4.0-compatible service |
+| MONGODB_MAX_NUMBER_OF_RETRIES  | 1     | Meadowlark can be configured to retry on MongoDB failure, for example when two concurrent transactions attempt to access the same resource. By default, Meadowlark will retry once. This can be changed with the MAX_NUMBER_OF_RETRIES environment variable. |
```
stephenfuqua, line 58:
```suggestion
| MONGODB_MAX_NUMBER_OF_RETRIES  | 1     | Meadowlark can be configured to retry on MongoDB failure, for example when two concurrent transactions attempt to access the same resource. By default, Meadowlark will retry once. This can be changed with the `MONGODB_MAX_NUMBER_OF_RETRIES` environment variable. |
```

stephenfuqua (approved):

stephenfuqua (approved):

andonyns (approved):

PR #218, opened 2023-03-09
[RND-490] Mongo memory resources are not being cleared after running tests

andonyns:
Improvements to mongo memory. This only runs when running integration tests.

There's one remaining issue with mongo memory when tests are stopped with ctrl + c and tests have not finished, where the mongo memory files in: C:\Users\\{user}\AppData\Local\Temp are not cleared
image
These files must be manually deleted.

This is an issue with Jest, see: https://github.com/facebook/jest/issues/5391 and https://github.com/facebook/jest/issues/6029 for reference


Meadowlark-js/.vscode/settings.json
```diff
@@ -98,5 +98,9 @@
       "markdown",
       "batch"
     ]
-  }
+  },
+  "jest.autoRun": {
+    "watch": false
+  },
+  "jest.jestCommandLine": "\"./node_modules/.bin/jest\" --runInBand",
```
andonyns, line 105: This allows the Jest extension to run all tests at the same time without having conflicts

Meadowlark-js/jest-mongodb-config.js
```diff
@@ -1,11 +1,11 @@
 module.exports = {
   mongodbMemoryServerOptions: {
     binary: {
-      version: '4.0.28',
+      version: '4.4.0',
```
stephenfuqua, line 4: This bothered me a bit at first, because testing against a newer version doesn't prove that we're only using 4.0-compatible code. However, so long as our e2e tests are using Mongo 4.0, then we're in good shape.

stephenfuqua (approved):

PR #220, opened 2023-03-14
[RND-510] Opensearch e2e testcontainer flakiness

jleiva-gap:
- Separate jest configuration files are created for each of the backends in the integration tests.
- Add javascript modules to setup and tear down the opensearch test environment.
- The package.json is updated to add individual integration test execution options for each of the backends.
- In the github action of the on-pullrequest, for the integration tests option, the --config parameter is removed, which is not necessary.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist


.github/workflows/on-pullrequest.yml
```diff
@@ -156,7 +156,7 @@ jobs:
             },
             {
               type: "Integration",
-              command: "npm run test:integration -- --ci --config ./tests/config/integration/jest.config.js",
+              command: "npm run test:integration",
```
stephenfuqua, line 159: I understand removing the config, but I think we still want `--ci`.

> `--ci`
> When this option is provided, Jest will assume it is running in a CI environment. This changes the behavior when a new snapshot is encountered. Instead of the regular behavior of storing a new snapshot automatically, it will fail the test and require Jest to be run with --updateSnapshot.

Meadowlark-js/backends/meadowlark-postgresql-backend/test/config/integration/jest.config.js
```diff
@@ -0,0 +1,18 @@
...
+  testMatch: ['**/meadowlark-postgresql-backend/test/integration/**/*.(spec|test).[jt]s?(x)'],
+  coverageThreshold: {
+    global: {
+      branches: 52,
+      functions: 58,
+      lines: 60,
+      statements: 60,
+    },
+  },
+  rootDir: '../../../../../',
```
andonyns, line 16: Given that we are inheriting from a default config, this might not be required

andonyns (approved):

PR #224, opened 2023-03-20
[RND-485] Publishing to Azure Artifacts

stephenfuqua:
Demonstration in [Stephen's fork](https://github.com/stephenfuqua/Meadowlark).

* [v0.3.3-pre-15](https://github.com/stephenfuqua/Meadowlark/releases/tag/v0.3.3-pre-15) was auto-generated _as a pre-release_ from a merge to `main`. It successfully published to Azure Artifacts.
* Next, I created [v0.3.4](https://github.com/stephenfuqua/Meadowlark/releases/tag/v0.3.4) _as a pre-release_
  * Must be created initially a pre-release, because that's when the build process occurs.
* When the `on-prerelease` workflow was done, edited v0.3.4 to change from _pre-release_ to _latest_ release, triggering the `on-release` workflow
* The latter promoted v0.3.4 to the "release view" in Azure Artifacts.

image

After this, I repeated the process to create a new pre-release after a pull request merge, and then created a release 0.3.5. 

https://dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_artifacts/feed/EdFi/Npm/@edfi%2Fmeadowlark-authz-server/versions/0.3.5


.github/workflows/on-release.yml
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+# Doesn't do anything yet
+
```
stephenfuqua, line 7:
```suggestion
```

That was temporary and should be removed

Meadowlark-js/backends/meadowlark-mongodb-backend/package.json
```diff
@@ -19,9 +19,9 @@
     "build:copy-non-ts": "copyfiles -u 1 -e \"**/*.ts\" \"src/**/*\" dist --verbose"
   },
   "dependencies": {
-    "@edfi/meadowlark-authz-server": "0.3.0-dev.0",
-    "@edfi/meadowlark-core": "0.3.0-dev.0",
-    "@edfi/meadowlark-utilities": "0.3.0-dev.0",
+    "@edfi/meadowlark-authz-server": "^v0.3.0-pre-35",
+    "@edfi/meadowlark-core": "^v0.3.0-pre-35",
+    "@edfi/meadowlark-utilities": "^v0.3.0-pre-35",
```
stephenfuqua, line 24: In all of my excitement to see if I have this process working, without committing changes to the package.json every time the version updates, I forgot to test and make sure this is actually working. The key question: if I make an update to the meadowlark-utilities, will the other packages pickup that change? Or are they going to try to pull pre-35 from Azure Artifacts and use it instead?

eng/promote-packages.ps1
```diff
@@ -0,0 +1,55 @@
...
+  `"views`": {
+    `"op`": `"add`",
+    `"path`": `"/views/-`",
+    `"value`": `"Release`"
+  }
+}"
+
+$messages = @()
+$packages | ForEach-Object {
+  $url = $urlTemplate.Replace("{PACKAGE}", "$($_.' $_ -split "/" | Select-Object -last 1 ')").Replace("{VERSION}", "$Version")
```
stephenfuqua, line 45: On line 26, I couldn't figure out how to get the `Select-Object` command to create a new object with a useful name. It automatically creates an object with the awkward name ` $_ -split "/" | Select-Object -last 1 `! And thus you see that above. If that doesn't quite make sense yet, here's an equivalent initialization statement in PowerShell:

```pwsh
$a = @{
  ' $_ -split "/" | Select-Object -last 1 ' = "meadowlark-utilities"
}
```

.github/workflows/on-prerelease.yml
```diff
@@ -0,0 +1,71 @@
...
+          echo //pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/npm/registry/:_password=$API_KEY >> $HOME/.npmrc
+          echo //pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/npm/registry/:email=techsupport@ed-fi.org >> $HOME/.npmrc
+          echo "; end auth token" >> $HOME/.npmrc
+
+      - name: Publish
+        run: npm publish --workspaces --tag dev
+
+      # NOTE: there are uncommitted version number changes in package.json. No
+      # problem! Our strategy is to rely on the tag history for version numbers,
+      # not on package.json.
```
stephenfuqua, line 71: @bradbanister this could end up coming back to haunt me.

In MetaEd, we have a GPG key for signing a commit and tag, which we can then push back to the repository. I wasn't thrilled about that process, but I accepted it. In this public repository I am even less enthusiastic about it, as there is more opportunity for a secret to spill out to the public... giving someone who manages to steal the GPG key full access to write commits into our repositories. Hence I'm hoping this version update will work even though it isn't being preserved in the package.json files.

bradbanister: Yes, I think the package.json versions getting out of sync will come back to haunt us.

It looks like in the past 6 months GitHub has added a fine-grained permissions option to access tokens: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#fine-grained-personal-access-tokens

Maybe we could use that to limit access to commit pushes in Meadowlark?

PR #230, opened 2023-04-05
[RND-512] e2e Developer Mode

andonyns:
Adding a developer mode for e2e tests to speed up the test creation/debugging process

## Description

TestContainers has a 'ryuk' container that runs as the handler of the container lifecycle.
Disabling this container and adding an option to reuse containers allows us to set up the container once and then use it for the tests.
To use we need to set the DEVELOPER_MODE flag and after running it once, save the Admin Key and Secret.

### DIFF
--- a/Meadowlark-js/package.json
+++ b/Meadowlark-js/package.json
@@ -60,7 +60,7 @@
     "test:e2e:build": "npm run docker:build && npm run test:e2e:jest",
     "test:e2e:build:wsl1": "npm run docker:build:wsl1 && npm run test:e2e:jest",
     "test:e2e:dev:setup": "cross-env TESTCONTAINERS_RYUK_DISABLED=true ts-node tests/e2e/setup/DevModeConfig.ts",
-    "test:e2e:dev:exit": "docker stop mongo-test opensearch-test meadowlark-api-test && docker rm -f mongo-test opensearch-test meadowlark-api-test",
+    "test:e2e:dev:exit": "docker stop mongo-test opensearch-test meadowlark-api-test && docker rm -f mongo-test opensearch-test meadowlark-api-test && docker network prune",
     "start:local": "lerna run start:local",
     "docker:lint": "cat ./Dockerfile | docker run --rm -i hadolint/hadolint",
     "docker:install": "npm i lerna@^6.0.2 rimraf@^3.0.2 copyfiles@^2.4.1 typescript@4.8.4 -g && npm ci --only=production",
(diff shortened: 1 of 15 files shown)


Meadowlark-js/package.json
```diff
@@ -59,6 +59,8 @@
     "test:e2e:debug": "cross-env DEBUG=testcontainers* npm run test:e2e:jest",
     "test:e2e:build": "npm run docker:build && npm run test:e2e:jest",
     "test:e2e:build:wsl1": "npm run docker:build:wsl1 && npm run test:e2e:jest",
+    "test:e2e:dev:setup": "cross-env TESTCONTAINERS_RYUK_DISABLED=true ts-node tests/e2e/setup/DevModeConfig.ts",
+    "test:e2e:dev:exit": "docker stop mongo-test opensearch-test meadowlark-api-test && docker rm -f mongo-test opensearch-test meadowlark-api-test && docker network prune -f",
```
andonyns, line 63: The network pruning only affects unused networks, if not done, there would be many docker networks stored without use

Meadowlark-js/package.json
```diff
@@ -59,6 +59,8 @@
...
     "start:local": "lerna run start:local",
```
stephenfuqua, line 64: Can we get something like this... ? And if I have the wrong command sequence, then that is all the more reason to help us with a one line version 😉 .

```suggestion
    "test:e2e:dev": "npm run test:e2e:dev:setup && npm run test:e2e:build && npm run test:e2e:exit",
    "start:local": "lerna run start:local",
```

stephenfuqua:
Can you add a note with instructions on running the e2e tests locally, in `docs/LOCALHOST.md`?

Meadowlark-js/package.json
andonyns: I understand, but I don't think this adds value given that this is doing the same as running the e2e tests without setting the DEVELOPMENT_MODE, since it's setting the containers, running the tests and then cleaning it. 
In that case, one could run the tests with the setting "turned off".

Do you have an scenario in mind?

docs/LOCALHOST.md
```diff
@@ -71,3 +71,36 @@ delete from meadowlark.references;
...
+If you're adding new features to Meadowlark, and want to test the changes in an isolated environment, run `npm run test:e2e:build`.
+This will build the Docker image and run the tests against that image, cleaning the environment afterwards.
+
+### Using published image
+
+If you want to run the tests against and existing Docker image, set the variable `API_IMAGE_NAME` in the .env file.
+Run the tests with `npm run test:e2e:jest`.
+This will pull the image and run the tests.
+
+### Developer Mode
```
andonyns, line 95: @stephenfuqua let me know if the details are clear.

I'm thinking if we should add an option to build just the backend in "dev-mode" that way we could then run the tests against an image that is being built constantly without clearing the backend. What are your thoughts?

stephenfuqua: I don't understand.

stephenfuqua (approved):

PR #235, opened 2023-04-18
[RND-494] Add postgres support in docker compose file

andonyns:
Add postgres configuration for docker compose file on root Meadowlark project

## Description

- Add postgres configuration with custom name to run API against postgres.
- Restore e2e tests
- Add e2e test setup for postgres

### DIFF
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
@@ -30,13 +30,15 @@ export async function deleteDocumentByDocumentUuid(
     await client.query('BEGIN');
     // Find the alias ids for the document to be deleted
     const documentAliasIdsResult: QueryResult = await client.query(findAliasIdsForDocumentByDocumentUuidSql(documentUuid));
+    // Each row contains documentUuid and corresponding meadowlarkId (document_id),
+    // we just need the first row to return the document_id
     meadowlarkId =
       documentAliasIdsResult?.rowCount != null && documentAliasIdsResult.rowCount > 0
         ? documentAliasIdsResult.rows[0].document_id
         : ('' as MeadowlarkId);
     if (validateNoReferencesToDocument) {
       // All documents have alias ids. If no alias ids were found, the document doesn't exist
-      if (documentAliasIdsResult.rowCount == null || documentAliasIdsResult.rowCount === 0) {
+      if (meadowlarkId === '') {
         await client.query('ROLLBACK');
         Logger.debug(`${moduleName}.deleteDocumentByDocumentUuid: DocumentUuid ${documentUuid} does not exist`, traceId);
         deleteResult = { response: 'DELETE_FAILURE_NOT_EXISTS' };
@@ -44,7 +46,7 @@ export async function deleteDocumentByDocumentUuid(
       }
 
       // Extract from the query result
-      const documentAliasIds: string[] = documentAliasIdsResult.rows.map((ref) => ref.alias_id);
+      const documentAliasIds: MeadowlarkId[] = documentAliasIdsResult.rows.map((ref) => ref.alias_id);
 
       // Find any documents that reference this document, either it's own id or an alias
       const referenceResult: QueryResult<any> = await client.query(findReferencingMeadowlarkIdsSql(documentAliasIds));
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
@@ -31,7 +31,7 @@ export async function deleteDocumentByDocumentUuid(
     // Find the alias ids for the document to be deleted
     const documentAliasIdsResult: QueryResult = await client.query(findAliasIdsForDocumentByDocumentUuidSql(documentUuid));
     meadowlarkId =
-      documentAliasIdsResult?.rowCount == null && documentAliasIdsResult.rowCount > 0
+      documentAliasIdsResult?.rowCount != null && documentAliasIdsResult.rowCount > 0
         ? documentAliasIdsResult.rows[0].document_id
         : ('' as MeadowlarkId);
     if (validateNoReferencesToDocument) {
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
@@ -7,51 +7,49 @@ import { Logger } from '@edfi/meadowlark-utilities';
 import type { DeleteResult, DeleteRequest, BlockingDocument, MeadowlarkId } from '@edfi/meadowlark-core';
 import type { PoolClient, QueryResult } from 'pg';
 import {
-  deleteDocumentByIdSql,
+  deleteDocumentByDocumentUuIdSql,
   deleteOutboundReferencesOfDocumentSql,
   findReferringDocumentInfoForErrorReportingSql,
   deleteAliasesForDocumentSql,
-  findAliasIdsForDocumentSql,
-  findReferencingDocumentIdsSql,
+  findReferencingMeadowlarkIdsSql,
+  findAliasIdsForDocumentByDocumentUuidSql,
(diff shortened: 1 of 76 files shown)


Meadowlark-js/package.json
```diff
@@ -60,7 +60,7 @@
     "test:e2e:build": "npm run docker:build && npm run test:e2e:jest",
     "test:e2e:build:wsl1": "npm run docker:build:wsl1 && npm run test:e2e:jest",
     "test:e2e:dev:setup": "cross-env TESTCONTAINERS_RYUK_DISABLED=true ts-node tests/e2e/setup/DevModeConfig.ts",
-    "test:e2e:dev:exit": "docker stop mongo-test opensearch-test meadowlark-api-test && docker rm -f mongo-test opensearch-test meadowlark-api-test && docker network prune -f",
+    "test:e2e:dev:exit": "bash ../eng/exit-dev-containers.sh",
```
stephenfuqua, line 63: I'm curious, does this work for you in Windows? I assume you have WSL2. I still have WSL (1), can't remember why but something "broke" with 2. Running this command from PowerShell, bash just hangs and I have to kill it manually.

What do you think about changing to a PowerShell script, since PowerShell is available in GitHub Actions?

Meadowlark-js/tests/e2e/setup/EnvironmentConfig.ts
```diff
@@ -7,30 +7,39 @@ import { Network, StartedNetwork } from 'testcontainers';
...
     console.warn(
-      '⚠️ WARNING: Running in DEVELOPER MODE. Containers should be already be started, if not, setup with `test:e2e:dev:setup`⚠️',
+      '⚠️ WARNING: Skipping initialization. Containers should be already be started, if not, setup with `test:e2e:dev:setup`⚠️',
     );
   } else {
-    process.env.ALREADY_SET = 'true';
     console.info('-- Setting up containers --');
     await Promise.all([MongoContainer.setup(network), ApiContainer.setup(network), OpenSearchContainer.setup(network)]);
+
+    if (process.env.DOCUMENT_STORE_PLUGIN === '@edfi/meadowlark-postgresql-backend') {
```
stephenfuqua, line 39: I read this line and asked myself, why not a condition before starting MongoDB?

Might be worth adding a comment above line 37 that MongoDB always runs so that we can support authentication.

Meadowlark-js/package.json
andonyns: We can do it in PowerShell. This script is not required to be executed on GitHub Actions, this is only a convenience for local development so there's no issue on changing to PS

andonyns: @stephenfuqua we have the [init-docker-compose.sh](https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/main/eng/init-docker-compose.sh) script where we handle wsl1 and wsl2, would something like this work for you?

Or do you prefer to have it as a PS script? I'm okay either way since we currently have a combination of both in the repo.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -28,19 +28,21 @@ export async function deleteDocumentByDocumentUuid(
...
+    meadowlarkId =
+      documentAliasIdsResult?.rowCount != null && documentAliasIdsResult.rowCount > 0
+        ? documentAliasIdsResult.rows[0].document_id
+        : ('' as MeadowlarkId);
     if (validateNoReferencesToDocument) {
-      // Find the alias ids for the document to be deleted
-      const documentAliasIdsResult: QueryResult = await client.query(findAliasIdsForDocumentByDocumentUuidSql(documentUuid));
-
       // All documents have alias ids. If no alias ids were found, the document doesn't exist
       if (documentAliasIdsResult.rowCount == null || documentAliasIdsResult.rowCount === 0) {
```
stephenfuqua, line 39: This check is now on line 34 as well. Would it make sense to combine in some way? Might be just fine the way it is though.

Meadowlark-js/package.json
stephenfuqua: It still just hangs on my computer, with no messages, and I have to kill it from the Task Manager (Control-C doesn't help). Since this is primarily a local command, and locally we're primarily running Windows, please change to PowerShell.

Meadowlark-js/docker-compose.yml
```diff
@@ -188,6 +192,19 @@ services:
...
+  #   hostname: postgres
+  #   restart: always
+  #   environment:
+  #     POSTGRES_USER: ${POSTGRES_USER:-postgres}
+  #     POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-abcdefgh1!}
+  #   ports:
+  #     - ${POSTGRES_PORT:-5432}:5432
+  #   volumes:
+  #     - pgsql-ml-data:/var/lib/postgresql/data
+
```
stephenfuqua, line 207: Why is this commented out?

Meadowlark-js/tests/e2e/readme.md
```diff
@@ -20,7 +20,7 @@ This is a special mode to use when adding new e2e tests or in situations where i
 - Set the environment variable `DEVELOPER_MODE=true`
 - Run `npm run test:e2e:jest` to run the tests. Save the Admin Key and Secret (as specified in [.env.example](./setup/.env.example)) to be able to run the tests
 without cleaning the environment
-- When done, run: `npm run test:e2e:dev:exit` to clean the environment
+- When done, run: `npm run test:e2e:dev:exit` or execute the `exit-dev-containers.ps1` script to clean the environment
```
stephenfuqua, line 23: Ah, two options now 👍

eng/exit-dev-containers.ps1
```diff
@@ -0,0 +1,42 @@
+# # SPDX-License-Identifier: Apache-2.0
+# # Licensed to the Ed-Fi Alliance under one or more agreements.
+# # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# # See the LICENSE and NOTICES files in the project root for more information.
+
+#Requires -Version 7
+# $ErrorActionPreference = "Stop"
```
stephenfuqua, line 7: Can this be uncommented, or deleted?

eng/exit-dev-containers.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+$docker=docker
+
+if ($a -gt 2) {
+  # Running under WSL1, need to reference docker.exe instead of docker. Simple
+  # alias doesn't work, sadly. This variable helps in the context of this
+  # script, but need to use "wsl1" commands with npm.
+  $docker=docker.exe
+}
```
stephenfuqua, line 16: This is unnecessary when using PowerShell instead of Bash: `docker` command works just fine. Also `$docker` appears to be unused. I recommend simplifying the code below to just use `docker` everywhere.

Granted, this might be a problem if you start `pwsh` _from within a Bash prompt_ -  that is, PowerShell running on Linux. But there's a good work around - just stay in Bash at that point 😆.

eng/exit-dev-containers.sh
```diff
@@ -0,0 +1,22 @@
...
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+docker=docker
+if [[ $(grep Microsoft /proc/version) ]]; then
+# Running under WSL1, need to reference docker.exe instead of docker. Simple
+# alias doesn't work, sadly. This variable helps in the context of this
+# script, but need to use "wsl1" commands with npm.
+docker=docker.exe
```
stephenfuqua, line 14:
```suggestion
  docker=docker.exe
```

stephenfuqua: Tiny commit so that I can approve the PR for codeowners purposes, but it still requires someone else on the team to test and merge the PR.

stephenfuqua (approved):

Meadowlark-js/docker-compose.yml
andonyns: I added the option but left it commented to avoid doing the setup every time and leave it optional

jleiva-gap (approved):

PR #236, opened 2023-04-29
[RND-493] Implementation design for offline cascading deletes

stephenfuqua:
Note the new recommended extension in VS Code.

### DIFF
--- a/docs/design/offline-cascading-deletes/README.md
+++ b/docs/design/offline-cascading-deletes/README.md
@@ -1,0 +1,403 @@
+# Design for Offline Cascading Deletes
+
+## Overview
+
+The Ed-Fi ODS/API [supports cascading
+updates](https://techdocs.ed-fi.org/x/jgigBw) on
+some resources' natural key values out of the box, and allows for customization
+to support on all resources. In the Ed-Fi Data Standard version 3 and version 4,
+the supported resources are:
+
+* Class Period
+* Grade
+* GradebookEntry
+* Location
+* Section
+* Session
+* Student School Association
+* Student Section Association
+
+Cascading updates are not difficult in the ODS/API, because of its use of
+relational database management systems that have strong support for this kind of
+key change.
+
+The Meadowlark Ed-Fi API, on the other hand, uses document storage with no
+inherent referential integrity. Therefore, if cascading updates are to be
+implemented, they will require additional code-level support.
+
+## Primer on Meadowlark Referential Integrity
+
+Whether using [MongoDB](https://techdocs.ed-fi.org/x/yARqBw) or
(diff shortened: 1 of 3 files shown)


docs/design/offline-cascading-deletes/README.md
```diff
@@ -0,0 +1,414 @@
...
+```
+
+The problem needs to be resolved upstream of the events - before recording the
+change in the primary data store. The likely solution is to have the API modify
+the document _immediately on receipt_, enriching with a Unix timestamp property
+`lastModifiedAt`. So long as the server times are in sync, then the API should
+avoid writing an update if the stored value has a larger `lastModifiedAt`. To be
+on the safe side, use similar logic in the event handling. This is a variation
+of optimistic locking. It may be helpful to think of this timestamp as an
+"identity version number".
```
stephenfuqua, line 360: Shoutout to "identity version number" mentioned by Brad in a ticket comment.

docs/design/offline-cascading-deletes/README.md
```diff
@@ -0,0 +1,414 @@
...
+                  "newMeadowlarkId": <NEW VALUE>,
+                  "newIdentity": {
+                    "sKeyPart1": "a",
+                    "sKeyPart2": "new-value",
+                  }
+                },
+                "timestamp": <current Unix timestamp>,
+                "id": <UUID v1>,
+                "partitionKey": <UUID of changed document>
+              }
```
stephenfuqua, line 296: Note that this is only writing one record to the outbox. Because both the MongoDB and PostgreSQL update commands can operate on an entire set, the API application doesn't "know" which records are being modified. Therefore it can't write individual update records to the event stream.

docs/design/offline-cascading-deletes/README.md
```diff
@@ -0,0 +1,414 @@
...
+the `key-update` event and then takes action to fix dependencies. Naive approach:
+
+1. For each document that references the `event.documentUuid` ("child document"):
+   1. For each component of `event.newIdentity`: replace child document's value
+      for matching key.
+   2. Calculate the child document's new Meadowlark ID and update the document.
+   3. Create a `key-update` event _for the child document_, so that any
+      "grandchildren" of the original change will also receive the update.
+   4. Create a `document-update` event for the child document, for updating
+      OpenSearch (details forthcoming).
```
stephenfuqua, line 320: Probably don't need (4), because OpenSearch would be updated from the events triggered by the `documents` collection/table.

docs/design/offline-cascading-deletes/README.md
bradbanister: I like this. Easier that managing a global "next version" counter somewhere.

bradbanister (approved):
Looks good!

stevenarnold-ed-fi (approved):

PR #251, opened 2023-06-07
[RND-533] Technical design for get all performance benchmarking

andonyns:
Adding details to run performance tests for Meadowlark

### DIFF
--- a/docs/design/performance-benchmarking/README.md
+++ b/docs/design/performance-benchmarking/README.md
@@ -8,23 +8,15 @@ To do the performance tests to retrieve endpoints, use the
 ### Running Performance tests for Meadowlark
 
 - Configure Meadowlark and verify that it's running.
-- Load Sample data with the Invoke-LoadGrandBend or Invoke-LoadPartialGrandBend
-  scripts.
-- Verify that the data has been loaded correctly through the API or from the
-  database.
-- Clone the
-  [Suite3-Performance-Testing](https://github.com/Ed-Fi-Exchange-OSS/Suite-3-Performance-Testing)
-  repository.
-- Install [python](https://www.python.org/downloads/) and
-  [poetry](https://python-poetry.org/docs/#installation).
+- Load Sample data with the Invoke-LoadGrandBend or Invoke-LoadPartialGrandBend scripts.
+- Verify that the data has been loaded correctly through the API or from the database.
+- Clone the [Suite3-Performance-Testing](https://github.com/Ed-Fi-Exchange-OSS/Suite-3-Performance-Testing) repository.
+- Install [python](https://www.python.org/downloads/) and [poetry](https://python-poetry.org/docs/#installation).
 - Go to /src/edfi_paging_test folder and run `poetry install`.
-- Create a .env file based on
-  /src/edfi_paging_test/edfi_paging_test/.env.example.
-- Add the desired endpoints or comment line to get all (get all functionality
-  blocked by RND-577).
+- Create a .env file based on /src/edfi_paging_test/edfi_paging_test/.env.example.
+- Add the desired endpoints or comment line to get all (get all functionality blocked by RND-577).
 - Run `poetry run python edfi_paging_test`
-- Verify results located in the specified path or in the /out folder in CSV
-  format.
+- Verify results located in the specified path or in the /out folder in CSV format.
 
 ### Analyzing Performance Results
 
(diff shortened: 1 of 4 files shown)


docs/design/performance-benchmarking/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Get All
+
+To do the performance tests to retrieve endpoints, use the
+[Suite3-Performance-Testing](https://github.com/Ed-Fi-Exchange-OSS/Suite-3-Performance-Testing).
+
+### Running Performance tests for Meadowlark
+
+- Configure Meadowlark and verify that it's running.
+- Load Sample data with the Invoke-LoadGrandBend or Invoke-LoadPartialGrandBend scripts.
```
stephenfuqua, line 11: Do we need a script for loading Southreidge?

andonyns: It can be good for convenience

docs/design/performance-benchmarking/README.md
```diff
@@ -0,0 +1,81 @@
...
+The generated csv files display the results of the execution.
+
+Results example:
+
+| Resource |                         URL                         | PageNumber | PageSize | NumberOfRecords | ElapsedTime | StatusCode |
+|----------|-----------------------------------------------------|------------|----------|-----------------|-------------|------------|
+| accounts | http://{meadowlark_url}/accounts?offset=0&limit=100 |  1         |    100   |       100       | 0.020013055 |     200    |
+| accounts | http://{meadowlark_url}/accounts?offset=0&limit=100 |  2         |    100   |       100       | 0.040413055 |     200    |
+| accounts | http://{meadowlark_url}/accounts?offset=0&limit=100 |  1         |    100   |       100       | 0.089013055 |     200    |
+
```
stephenfuqua, line 32: Let's put some more definition around the process. Looking at the individual timings is potentially useful when debugging, but for "get all" the individual times for the resources shouldn't matter very much. Looking at the total amount of time taken for all resources will be interesting.

So let's make sure that we can easily record the total time taken to execute the `poetry run python edfi_paging_test` command. If there is not already a value displayed by the code, then we can use PowerShell's `Measure-Command` to get the timing.

Then we need to generate statistics, because one time is not enough. Best to run at least five times so that we can get an average and a meaningful standard deviation. To be honest, 30 times would be best, but this is going to take too long (30 times is the rule of thumb for the law of large numbers, making it very likely that you will have a bell curve to your data).

andonyns: The command shows the total execution time, I will add this to the details.

When you mention about generating statistics, do you mean for me to:
1) Do the multiple runs to show the actual values I got from my run? The values shown here are a small example with the values I got from my run, I can add the entire table for it to be a "real world" example.
2) Document the steps to do so?
3) Generate a script that does that?

I like the idea of generating a script but I'm not sure if this is expected as part of this spike or if we should add a follow up ticket to do so.

stephenfuqua: I added the response above without reloading the page, thus had not yet seen your response, Andony.

The table above is fine. But it does point to timings for individual resources, i.e. for `accounts`. I'm talking about timing the entire execution of the the paging test. 

```powershell
$numTrials = 5
$times = @()

for ($i = 0; $i -lt $numTrials; $i++) {
    $timing = Measure-Command { sleep 1 }
    $times += $timing.TotalMilliseconds
}

$sum = 0.0
$times | ForEach-Object { $sum += $_ }
$mean= $sum / $numTrials

$sumSquareError = 0.0
$times | ForEach-Object { $sumSquareError += [Math]::Pow($_ - $mean, 2) }
$standardDeviation = [Math]::Sqrt($sumSquareError / $numTrials)

Write-Output @"
Mean: $mean
Standard Deviation: $standardDeviation
"@
```

Naturally, replace `sleep 1` with `poetry run python edfi_paging_test`. I left the sleep command in because that's how I tested the script.

Anytime we make a change and run a test, the two critical numbers to report are the mean and standard deviation.

andonyns: @stephenfuqua I added the script and updated the documentation

docs/design/read-performance-benchmarking/README.md
```diff
@@ -0,0 +1,124 @@
...
+| accounts |   100    |       2       |         5       |   0.1031  | 0.020013055 |    0.0215   |        0       |
+| accounts |   100    |       2       |         5       |   0.1031  | 0.020013055 |    0.0215   |        0       |
+| accounts |   100    |       2       |         5       |   0.1031  | 0.020013055 |    0.0215   |        0       |
+
+### Comparing with ODS/API
+
+Run the same tests against and ODS/API instance with the same data set and filtering out the tpdm, sample and homograph
+resources since those are not handled by Meadowlark. Changing the url and variables in the .env file inside edfi_paging_test.
+
+### Profiler
```
stephenfuqua, line 80: Fine to leave this about the profiler here for future reference, but don't forget the paging tests will be hitting OpenSearch rather than MongoDB (except for authentication).

PR #258, opened 2023-06-14
[BIA-563] PostgreSQL backend SqlHelper functions should directly return typed results rather than strings

jleiva-gap:
Update sqlHelper to execute sql queries and return an object with the result. Update references to use the new result instead of execute the sql string. Update tests.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkAlias.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkAlias.ts
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+import { MeadowlarkId } from '@edfi/meadowlark-core';
+import { MeadowlarkDocumentId } from './MeadowlarkDocument';
+
+export interface MeadowlarkAlias extends MeadowlarkDocumentId {
+  /**
+   * Alias meadowlarkId.
+   */
+  alias_meadowlark_id: MeadowlarkId;
+}
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkDocument.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkDocument.ts
@@ -3,7 +3,7 @@
 // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 // See the LICENSE and NOTICES files in the project root for more information.
 
-import { DocumentIdentity, MeadowlarkId, DocumentUuid } from '@edfi/meadowlark-core';
+import { DocumentIdentity, MeadowlarkId, NoDocumentIdentity, DocumentUuid } from '@edfi/meadowlark-core';
 
 export interface MeadowlarkDocumentId {
   /**
@@ -61,10 +61,31 @@ export interface MeadowlarkDocument extends MeadowlarkDocumentId {
   created_by: string;
 }
 
-export function getEmptyMeadowlarkDocument(): MeadowlarkDocument {
-  return {} as MeadowlarkDocument;
+/**
+ * Creates a new empty newMeadowlarkDocument object
+ */
+export function newMeadowlarkDocument(): MeadowlarkDocument {
+  return {
+    meadowlark_id: undefined as unknown as MeadowlarkId,
+    document_uuid: undefined as unknown as DocumentUuid,
+    document_identity: NoDocumentIdentity,
+    project_name: '',
+    resource_name: '',
+    resource_version: '',
+    edfi_doc: null,
+    validated: false,
+    is_descriptor: false,
+    created_by: '',
+  };
 }
 
+/**
+ * The null object for DocumentInfo
+ */
+export const NoMeadowlarkDocument = Object.freeze({
+  ...newMeadowlarkDocument(),
+});
+
 export function isMeadowlarkDocumentEmpty(meadowlarkDocument: MeadowlarkDocument): boolean {
-  return Object.keys(meadowlarkDocument).length === 0;
+  return meadowlarkDocument === NoMeadowlarkDocument;
 }
--- a/Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkDocument.ts
+++ b/Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkDocument.ts
@@ -1,0 +1,70 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+import { DocumentIdentity, MeadowlarkId, DocumentUuid } from '@edfi/meadowlark-core';
+
+export interface MeadowlarkDocumentId {
+  /**
+   * A string hash derived from the project name, resource name
+   * and identity of the API document. This field replaces the built-in MongoDB _id.
+   */
+  meadowlark_id: MeadowlarkId;
+  /**
+   * The UUID for the document.
+   */
+  document_uuid: DocumentUuid;
+}
+
+export interface MeadowlarkDocument extends MeadowlarkDocumentId {
+  /**
+   * The identity elements extracted from the API document.
+   */
+  document_identity: DocumentIdentity;
+
+  /**
+   * The MetaEd project name the API document resource is defined in e.g. "EdFi" for a data standard entity.
+   */
+  project_name: string;
+
+  /**
+   * The name of the resource. Typically, this is the same as the corresponding MetaEd entity name. However,
+   * there are exceptions, for example descriptors have a "Descriptor" suffix on their resource name.
+   */
+  resource_name: string;
+
+  /**
+   * The resource version as a string. This is the same as the MetaEd project version
+   * the entity is defined in e.g. "3.3.1-b" for a 3.3b data standard entity.
+   */
+  resource_version: string;
+
+  /**
+   * The Ed-Fi ODS/API document itself.
+   */
+  edfi_doc: any;
+
+  /**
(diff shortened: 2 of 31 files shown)


Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -48,22 +48,20 @@ export async function deleteDocumentByDocumentUuid(
...
+      const documentAliasMeadowlarkIds: MeadowlarkId[] = documentAliasIdsResult.map((ref) => ref.alias_meadowlark_id);
 
       // Find any documents that reference this document, either it's own meadowlarkId or an alias
-      const referenceResult: QueryResult<any> = await client.query(
-        findReferencingMeadowlarkIdsSql(documentAliasMeadowlarkIds),
-      );
+      const referenceResult: MeadowlarkId[] = await findReferencingMeadowlarkIdsSql(client, documentAliasMeadowlarkIds);
 
-      if (referenceResult.rows == null) {
+      if (referenceResult == null) {
```
bradbanister, line 56: This should be a check for empty array. `findReferencingMeadowlarkIdsSql` never returns `null`

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
```diff
@@ -2,163 +2,267 @@
...
+ * Returns a list of meadowlarkIds of documents that reference the given meadowlarkIds.
  *
- * @param referencedMeadowlarkIds the meadowlarkIds of the documents that are being referenced
- * @returns a SQL query to find the meadowlarkIds of the documents referencing the given meadowlarkIds
+ * @param referencedMeadowlarkIds the array of meadowlarkIds of the documents that are being referenced
+ * @returns an array of meadowlarkIds of the documents referencing the given meadowlarkIds
  */
-export function findReferencingMeadowlarkIdsSql(referencedMeadowlarkIds: MeadowlarkId[]): string {
-  return format(
+export async function findReferencingMeadowlarkIdsSql(
```
bradbanister, line 17: Drop `Sql` from the names

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -29,15 +30,14 @@ export async function deleteDocumentByDocumentUuid(
   try {
     await client.query('BEGIN');
```
bradbanister, line 31: Suggest pushing this down to `SqlHelper` also. I don't think Delete/Read/Update/Upsert.ts should ever call `client.query()`. Leave that to the `SqlHelper` layer.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkDocument.ts
```diff
@@ -0,0 +1,70 @@
...
+  is_descriptor: boolean;
+
+  /**
+   * Creator of this document
+   */
+  created_by: string;
+}
+
+export function getEmptyMeadowlarkDocument(): MeadowlarkDocument {
+  return {} as MeadowlarkDocument;
```
bradbanister, line 65: This cast is toxic!!!  `{}` is in no way a `MeadowlarkDocument` and will crash any code that tries to use it as such.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkDocument.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function isMeadowlarkDocumentEmpty(meadowlarkDocument: MeadowlarkDocument): boolean {
+  return Object.keys(meadowlarkDocument).length === 0;
```
bradbanister, line 69: This is no good either. Because of the above, this function is effectively **required** before any use of a `MeadowlarkDocument`.

Instead, use the Null Object pattern and introduce a singleton null object called `NoMeadowlarkDocument`. This is done in many places in the Meadowlark codebase. A good example is https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/main/Meadowlark-js/packages/meadowlark-core/src/model/DocumentInfo.ts

Because it's a singleton, you can check for `=== NoMeadowlarkDocument` if you need to know.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -27,43 +31,32 @@ export async function deleteDocumentByDocumentUuid(
...
-      const documentAliasMeadowlarkIds: MeadowlarkId[] = documentAliasIdsResult.rows.map((ref) => ref.alias_meadowlark_id);
+      const documentAliasMeadowlarkIds: MeadowlarkId[] = documentAliasIdsResult.map((ref) => ref.alias_meadowlark_id);
 
       // Find any documents that reference this document, either it's own meadowlarkId or an alias
-      const referenceResult: QueryResult<any> = await client.query(
-        findReferencingMeadowlarkIdsSql(documentAliasMeadowlarkIds),
-      );
-
-      if (referenceResult.rows == null) {
-        await client.query('ROLLBACK');
```
bradbanister, line 59: Can we really remove this? Or does it need to go into `findReferencingMeadowlarkIdsSql` and throw an exception with the error message?

jleiva-gap: If we move that to the helper, should we have validation for all queries? Because at this moment the nulls are validated to return the empty object, but it could be that for other cases the client returns null the result.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
```diff
@@ -2,163 +2,269 @@
...
+export async function findReferencingMeadowlarkIds(
+  client: PoolClient,
+  referencedMeadowlarkIds: MeadowlarkId[],
+): Promise<MeadowlarkId[]> {
+  const querySelect = format(
     `SELECT parent_meadowlark_id FROM meadowlark.references WHERE referenced_meadowlark_id IN (%L)`,
     referencedMeadowlarkIds,
   );
+  const queryResult: QueryResult<any> = await client.query(querySelect);
+  return ((queryResult?.rowCount ?? 0) > 0 ? queryResult.rows.map((ref) => ref.parent_meadowlark_id) : []) as MeadowlarkId[];
```
bradbanister, line 51: Since `(queryResult?.rowCount ?? 0) > 0` is ugly, non-obvious, and repeated 11 times, let's extract it to a function like `function hasResults(queryResult: QueryResult<any>)`

bradbanister: There are still a bunch of functions in `SqlHelper` that return SQL strings that are being used by Update and Upsert. 😮 The goal should be for `client.query()` to **never** be called outside of `SqlHelper`.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
bradbanister: Got it. It's fine as is.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/Delete.ts
```diff
@@ -27,43 +31,32 @@ export async function deleteDocumentByDocumentUuid(
...
+      client,
+      documentUuid,
     );
     // Each row contains documentUuid and corresponding meadowlarkId (meadowlark_id),
     // we just need the first row to return the meadowlark_id
     meadowlarkId =
-      documentAliasIdsResult?.rowCount != null && documentAliasIdsResult.rowCount > 0
-        ? documentAliasIdsResult.rows[0].meadowlark_id
-        : ('' as MeadowlarkId);
+      (documentAliasIdsResult?.length ?? 0) > 0 ? documentAliasIdsResult[0].meadowlark_id : ('' as MeadowlarkId);
```
bradbanister, line 43: You can simplify `(documentAliasIdsResult?.length ?? 0) > 0` down to `documentAliasIdsResult.length > 0` when you are outside SqlHelper. You are already promising it will return an array in the type signature, and you have control over that.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/model/MeadowlarkAlias.ts
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+import { MeadowlarkId } from '@edfi/meadowlark-core';
+import { MeadowlarkDocumentId } from './MeadowlarkDocument';
+
+export interface MeadowlarkAlias extends MeadowlarkDocumentId {
```
bradbanister, line 9: I'd rename this to `MeadowlarkDocumentAndAliasId`, since that's what it is.

bradbanister:
Go through SqlHelper.ts and make sure the function documentation is updated. For example, I see several functions that say they return SQL strings that do not.

Meadowlark-js/backends/meadowlark-postgresql-backend/src/repository/SqlHelper.ts
```diff
@@ -169,13 +289,21 @@ export function findReferringDocumentInfoForErrorReportingSql(referringMeadowlar
  * @param aliasId the alias meadowlarkId for the given document, this may be the same as the meadowlarkId
  * @returns SQL query string to insert into the aliases table
  */
-export function insertAliasSql(documentUuid: DocumentUuid, meadowlarkId: MeadowlarkId, aliasId: MeadowlarkId): string {
-  return format(
+export async function executeInsertAlias(
```
bradbanister, line 292: Not a strong feeling here, but consider whether the `execute` prefix on these fuinction names are worthwhile.

stephenfuqua: When I click on "Files changed" tab and search for "function executeInsertAlias" I cannot find this function at all. I wanted to review the content of the function before commenting. This seems rather odd to me.

bradbanister (approved):
Looks good!

PR #268, opened 2023-07-18
[RND-537] Resource Caching Design

stephenfuqua:
work-in-progress


docs/design/caching/README.md
```diff
@@ -0,0 +1,398 @@
...
+        participant F as Backend.ReferenceValidation
+    end
+
+    participant E as [Transactional DB]
+
+    B->>C: upsertDocument(request, cache)
+    C->>D: upsertDocument(request, cache, client)
+    D->>D: startTransaction()
+
+    rect ivory
```
andonyns, line 247: There are a few rectangles that have a specific color (I guess that's from the *ivory* definition here) that do not look correctly on dark themes from GitHub, we should use another color to avoid this cases. In my case, the dark theme is my default even as incognito an makes it difficult to understand what's happening.

image

stephenfuqua: Yes, that is hard to read. I don't know how to change the text color. The ivory was used to highlight what has changed. Maybe I can find a way to make the entire diagram have a white background even when in a dark theme.

andonyns (approved):

PR #291, opened 2023-08-22
[RND-521] - Changes from the current SQL style query to DSL query on Opensearch.

DavidJGapCR:
## Types of changes

## Checklist

- [ ] I have signed all of the commits on this PR.
- [ ] I have agreed to the Ed-Fi Individual Contributors' License
- [ ] My code follows the code style of this project.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] All new and existing tests passed.

### DIFF
--- a/Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/QueryOpensearch.ts
+++ b/Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/QueryOpensearch.ts
@@ -22,44 +22,19 @@ export function indexFromResourceInfo(resourceInfo: ResourceInfo): string {
 }
 
 /**
- * Sanitizes input by escaping single all apostrophes. If \' is present, then someone is trying to evade the sanitization. In
- * that case, replace the search term with an empty string.
+ * DSL querying
  */
-function sanitize(input: string): string {
-  if (input.indexOf(`\\'`) > -1) {
-    return '';
-  }
-
-  return input.replaceAll(`'`, `\\'`);
-}
-
-/**
- * Convert query string parameters from http request to OpenSearch
- * SQL WHERE conditions. Returns empty string if there are none.
- */
-function whereConditionsFrom(queryParameters: object): string {
-  return Object.entries(queryParameters)
-    .map(([field, value]) => `${field} = '${sanitize(value)}'`)
-    .join(' AND ');
-}
-
-/**
- * This mechanism of SQL querying is specific to OpenSearch (vs Elasticsearch)
- */
-async function performSqlQuery(client: Client, query: string): Promise<any> {
+async function performDslQuery(client: Client, path: string, query: any, size: number, from: number): Promise<any> {
   return client.transport.request({
     method: 'POST',
-    path: '/_opendistro/_sql',
-    body: { query },
+    path: `/${path}/_search`,
+    body: {
+      from,
+      query,
+      size,
+      sort: [{ _doc: { order: 'asc' } }],
+    },
   });
-}
-
-/**
- * Returns well-formed WHERE clause from existing where clause and new clause
- */
-function appendedWhereClause(existingWhereClause: string, newWhereClause: string): string {
-  if (existingWhereClause === '') return newWhereClause;
-  return `${existingWhereClause} AND ${newWhereClause}`;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/QueryOpensearch.ts
```diff
@@ -73,35 +48,49 @@ export async function queryDocuments(request: QueryRequest, client: Client): Pro
...
-    let whereClause: string = '';
+    const matches: any[] = [];
 
     // API client requested filters
     if (Object.entries(queryParameters).length > 0) {
-      whereClause = whereConditionsFrom(queryParameters);
+      Object.entries(queryParameters).forEach(([key, value]) => {
+        matches.push({
+          match_phrase: { [key]: value },
+        });
```
stephenfuqua, line 58: I did some searching for equivalent to "SQL injection" and didn't find a lot. [This Stack Overflow](https://stackoverflow.com/questions/54769528/preventing-nosql-injections-with-elasticsearch) is interesting.

[Another post](https://stackoverflow.com/questions/64169248/elasticsearch-code-injection-when-using-c-sharp-nest-client) has some guidance that is interesting but not very helpful.

The main thing is to try to think: is there some way that the `key` or `value` could run an OpenSearch command? I just don't see that as likely at this point. Especially since we'll have validation on the `key` itself (because of the query string validator in Meadowlark). The value, if it is a string, will be appropriately quoted.

So I think I'm convinced that we have the right approach here and need not be worried about any other layers of protection.

bradbanister:
@DavidJGapCR I might be missing it, but do we have explicit tests for Opensearch and Elasticsearch showing that querying the same thing with different casing works? Those would be great to have, both as test and as test-as-documentation of that feature.

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
```diff
@@ -34,6 +34,30 @@ describe('When retrieving information', () => {
...
+          .then((response) => {
+            expect(response.headers['total-count']).toEqual('1');
+          });
+      });
+    });
+
+    describe('when querying changing the casing of the query value to be all uppercase', () => {
+      it('should return correct result', async () => {
+        await baseURLRequest()
+          .get(`/v3.3b/ed-fi/schools?nameOfInstitution=NEW SCHOOL 1`)
```
stephenfuqua, line 52: @bradbanister doesn't this cover the casing test? "new school 1" above and "NEW SCHOOL 1" here.

DavidJGapCR:
> @DavidJGapCR I might be missing it, but do we have explicit tests for Opensearch and Elasticsearch showing that querying the same thing with different casing works? Those would be great to have, both as tests and as test-as-documentation of that feature.

Not sure what you mean with test-as-documentation. 
If you see I added 2 tests to the e2e tests section. Those tests are running for all the combinations: mongo with opensearch, mongo with elastic search, pgsql with opensearch and pgsql with elasticsearch. This is in my last commit: https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/pull/291/commits/9ead010ab88a35b5d287c9012753d23c0d5d2b61

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
bradbanister: I was thinking in the actual `Meadowlark-js/backends/meadowlark-elasticsearch-backend/test/QueryElasticsearch.test.ts` tests. Something like `describe('when querying with a case different from the case in the datastore')`

bradbanister: That explains quite clearly that case-independent search is supported

Meadowlark-js/backends/meadowlark-opensearch-backend/src/repository/QueryOpensearch.ts
bradbanister: Also, to be blunt the e2e tests are not very helpful to me during development. Half the time I can't get them to run locally without container configuration errors that need lots of manual intervention.

Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts
bradbanister: Also, to be blunt the e2e tests are not very helpful to me during development. Half the time I can't get them to run locally without container configuration errors that need lots of manual intervention.

jleiva-gap (approved):

PR #318, opened 2023-11-13
[RND-649] Add lerna version update process into GitHub Action

andonyns:
Use lerna update process to apply changes into packages and commit changes into code.

Description
Delegate the version changes calculation to Lerna (without creating a commit).
Use new action to commit changes as a verified user.


Meadowlark-js/lerna.json
```diff
@@ -3,12 +3,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "v0.3.0-pre-35",
+  "version": "0.4.0-pre.0",
   "npmClient": "npm",
-  "useWorkspaces": true,
-  "command": {
```
andonyns, line 9: Although this is recommended to avoid tagging code in a different branch, in this case it is safe because the github action will not push code but only do the changes without commit

stephenfuqua:
@andonyns can you test this on a fork, with branch protection configured on `main` (requiring that there is a pull request in order to merge into `main`).

PR #320, opened 2023-11-14
[RND-649] Add lerna version update process into GitHub Action

andonyns:
Use lerna update process to apply changes into packages and commit changes into code.

Description
- Delegate the version changes calculation to Lerna (without creating a commit).
- Use new action to commit changes as a verified use
- Avoid multiple version updates from PRs
- Update node typing, versions and documentation where needed

To see the changes in place, see https://github.com/andonyns/Meadowlark
Main branch has write protection so all changes are done via PR.
See https://github.com/andonyns/Meadowlark/pull/18 as a demo of a PR ready to be merged and tag has not been created


Meadowlark-js/lerna.json
```diff
@@ -3,12 +3,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "v0.3.0-pre-35",
+  "version": "0.4.0-pre.0",
   "npmClient": "npm",
-  "useWorkspaces": true,
-  "command": {
```
andonyns, line 9: This could potentially allow users to create tags with lerna from branches, but this is currently possible from git tags also

Meadowlark-js/lerna.json
```diff
@@ -3,12 +3,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "v0.3.0-pre-35",
+  "version": "0.4.0-pre.0",
```
andonyns, line 6: This update had to be done manually for the first version, this will be done automatically in future changes

.github/workflows/on-pullrequest.yml
```diff
@@ -112,6 +116,49 @@ jobs:
...
+    steps:
+      - name: Checkout the Repo
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+        with:
+          fetch-depth: 0w
+
+      - name: Get changes
+        id: changes
+        run: |
+          suggested=v$(cat lerna.json | jq -r .version)
```
andonyns, line 138: To check weather or not we need to add a new commit for the new version, we compare the value in lerna.json against the latest tag available, if it's different, we assume that a change has already be done as part of this PR.

**warning** this means that manually updating the lerna.json version (not recommended) will cause the update not to run

.github/workflows/on-pullrequest.yml
```diff
@@ -12,6 +12,10 @@ on:
       - ".github/**"
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
stephenfuqua, line 17: 👍

.github/workflows/on-pullrequest.yml
stephenfuqua: @bradbanister FYI

stephenfuqua (approved):

PR #325, opened 2023-11-16
[RND-671] Fix Meadowlark CodeQL scan

andonyns:
- Update CodeQL action versions.
- Reorganize on pull request action to update dependencies before analyzing the code because there could be sync issues between the two.
- Adding Coverage flags to unit tests run.


.github/workflows/on-pullrequest.yml
```diff
@@ -29,10 +29,55 @@ jobs:
...
+
+      - uses: planetscale/ghcommit-action@4131649dbf2fdf1eb34421702972a5af7b0a8731 #v0.1.18
+        if: ${{ ! steps.changes.outputs.changes }}
+        with:
+          commit_message: "${{steps.set-version.outputs.version}}"
+          repo: ${{ github.repository }}
+          branch: ${{ github.head_ref || github.ref_name }}
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
```
stephenfuqua, line 75: I thought we already merged this work?

.github/workflows/on-pullrequest.yml
```diff
@@ -29,10 +29,55 @@ jobs:
...
   analyze-dependencies:
     # This Action will fail if run on anything other than a pull request
     if: github.event_name == 'pull_request'
     name: Analyze Code Dependencies
+    needs: upgrade
```
stephenfuqua, line 80: Would it make sense to add a comment explaining why `upgrade` must come first? I'm assuming because, otherwise, the analysis will run against the last version of the packages, instead of the new (current branch) version.

.github/workflows/on-pullrequest.yml
stephenfuqua: NVM, I see it was just a reorganization to move this higher up, since other jobs now depend on it.

andonyns:
See the workflow change in action here: https://github.com/andonyns/Meadowlark/actions/runs/6896275766

stephenfuqua (approved):

PR #326, opened 2023-11-21
[RND-653] Refactor Config file to not require mongodb URI

andonyns:
- Setting up mongo in e2e tests only when required.
- Update e2e tests documentation.
- Refactor versioning logic to run tests in main without adding multiple ifs in the code.

### DIFF
--- a/.github/workflows/on-pullrequest.yml
+++ b/.github/workflows/on-pullrequest.yml
@@ -30,9 +30,7 @@ jobs:
     uses: ed-fi-alliance-oss/ed-fi-actions/.github/workflows/repository-scanner.yml@main
 
   upgrade:
-    # Upgrade packages on PR only to avoid a double update when pushed to main
     name: Upgrade packages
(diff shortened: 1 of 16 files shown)


.github/workflows/on-pullrequest.yml
```diff
@@ -45,27 +43,29 @@ jobs:
         with:
           fetch-depth: 0w
 
-      - name: Get changes
-        id: changes
+      - name: Get version changes
+        id: versions
```
andonyns, line 47: This is a refactor of the logic to check if it's a PR in the script, that way we avoid having to add the always() logic in the subsequent jobs, instead of line 35

andonyns: See it in action here:
on main:
https://github.com/andonyns/Meadowlark/actions/runs/6947706833
on PR:
https://github.com/andonyns/Meadowlark/actions/runs/6947715142

stephenfuqua (approved):

PR #331, opened 2023-12-12
[RND-665] Lower rate limiting for OAuth

andonyns:
Setup and documentation for doing rate limiting with nginx setup

### DIFF
--- a/Meadowlark-js/sample-load-balancer.conf
+++ b/Meadowlark-js/sample-load-balancer.conf
@@ -1,3 +1,5 @@
+limit_req_zone $binary_remote_addr zone=oauthlimit:10m rate=1r/m;
+
 upstream meadowlarkoauth {
     server meadowlark-oauth-ml-local-1:3000;
     server meadowlark-oauth-ml-local-2:3000;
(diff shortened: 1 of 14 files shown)


Meadowlark-js/sample-load-balancer.conf
```diff
@@ -34,6 +36,12 @@ server {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 
+    location /local/oauth/token {
+        proxy_pass http://meadowlarkoauth;
+        
+        limit_req zone=oauthlimit burst=3 nodelay;
```
stephenfuqua, line 42: > "By default NGINX responds with status code 503 (Service Temporarily Unavailable) when a client exceeds its rate limit. Use the [limit_req_status](http://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_status) directive to set a different status code (444 in this example):
```
location /login/ {
    limit_req zone=mylimit burst=20 nodelay;
    limit_req_status 444;
}```

From https://www.nginx.com/blog/rate-limiting-nginx/

That's a weird status code to use. I would have expected status [429](https://www.webfx.com/web-development/glossary/http-status-codes/what-is-a-429-status-code/). 

I tried changing to 429. Worked well. But you provided a link to this same guidance, and anyone who prefers 429 can easily switch. This is just a demonstration.

stephenfuqua (approved):
Works well and is a good approach.

PR #337, opened 2024-01-08
[RND-676] Kubernetes Configuration for Meadowlark

andonyns:
Add a set of configuration files to run Meadowlark in a Kubernetes cluster

## Description

- Migrate from docker compose using komposer.
- Create ConfigMap to handle parameters, and Secret file to handle passwords
- Add instructions to deploy locally.

### DIFF
--- a/k8s/README.md
+++ b/k8s/README.md
@@ -1,0 +1,93 @@
+# Meadowlark Kubernetes Deployment
+
+This folder provides a basic setup of a set of
+[Kubernetes](https://kubernetes.io/) files to setup a cluster.
+
+## Local Development
+
+For local development, you need to use
+[minikube](https://minikube.sigs.k8s.io/docs/start/).
+
+* After installing, run `minikube start` to setup minikube in your local
+  environment.
+* Set the terminal in the */k8s* folder.
+* Run `kubectl apply -f .` to apply all or go file by file (`kubectl apply -f
+  {file-name}`).
+* After done, inspect with `kubectl get pods`, and verify that all pods have
+  status **RUNNING** (This can take a couple of minutes).
+
+This will start the kubernetes infrastructure to run without exposing any
+connection to the external network. When installing in a cloud provider the
+clouds Load Balancing service will take care of making the connection to the
+cluster, by opening a connection to the
+[meadowlark-api-service](meadowlark-api-service.yaml).
+
+This container has the type LoadBalancer, meaning that this is the entry point
+for the load balancer provider.
+
+To test this in the local environment, we need to open *tunnel* between the
+local network and the Kubernetes cluster. To do so, run `minikube service
+meadowlark-api --url`.
+
+Copy the URL and connect to Meadowlark.
+
+### View Logs
+
+To get the logs from Meadowlark, verify that the flag `LOG_TO_FILE` is set to
+true. If saving to the default location, get the logs by:
+
+1. Get the pod id `kubectl get pods`.
+2. Run `kubectl exec -it $POD_ID -- cat /var/log/meadowlark.log`
+
+### Useful commands
+
+| Command                                         | Description                  |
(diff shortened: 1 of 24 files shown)


k8s/app-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+    name: app-secret
+type: Opaque
+data:
+    postgres-password: dG9wc2VjdXJl
```
stephenfuqua, line 7: Is there an alternative to having a hard-coded secret here? Does this file truly belong in source control?

k8s/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Copy the URL and connect to Meadowlark.
+
+### View Logs
+
+To get the logs from Meadowlark, verify that the flag `LOG_TO_FILE` is set to
+true. If saving to the default location, get the logs by:
+
+1. Get the pod id `kubectl get pods`.
+2. Run kubectl exec -it $POD_ID -- cat /var/log/meadowlark.log
```
stephenfuqua, line 40:
```suggestion
2. Run `kubectl exec -it $POD_ID -- cat /var/log/meadowlark.log`
```

k8s/meadowlark-api-deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+            - name: ALLOW_OVERPOSTING
+              value: "false"
+            - name: ALLOW_TYPE_COERCION
+              value: "true"
+            - name: ALLOW__EXT_PROPERTY
+              value: "true"
+            - name: AUTHORIZATION_STORE_PLUGIN
+              value: '@edfi/meadowlark-postgresql-backend'
+            - name: DOCUMENT_STORE_PLUGIN
+              value: '@edfi/meadowlark-postgresql-backend'
```
stephenfuqua, line 30: Would it make sense to standardize on `'` or `"` rather than having both?

Do these value strings actually need to be delimited with one or the other? For example, docker compose doesn't need `"` around string values for the most part (maybe some exceptions though).

k8s/meadowlark-api-deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+                  key: opensearch-password
+            - name: OPENSEARCH_USERNAME
+              valueFrom:
+                configMapKeyRef:
+                  name: app-configmap
+                  key: opensearch-user
+            - name: OWN_OAUTH_CLIENT_ID_FOR_CLIENT_AUTH
+              value: meadowlark_verify-only_key_1
+            - name: OWN_OAUTH_CLIENT_SECRET_FOR_CLIENT_AUTH
+              value: meadowlark_verify-only_secret_1
```
stephenfuqua, line 78: Maybe we should move this to external configuration?

k8s/app-secret.yaml
andonyns: This is the equivalent of what we currently do in Docker Compose when setting the postgres password in the .env file, this is a minimal security configuration and will not actually be a secret, just a base64 encrypted script (in fact some blogs mention that the name secret is misleading) 

https://kubernetes.io/docs/concepts/configuration/secret/ shows a warning about "thinking" that the secrets are actually a security measure.

I think that for the purpose of the demo we could keep the file, but add a notice in the readme to use a different security method for passwords

k8s/meadowlark-api-deployment.yaml
andonyns: That's true, this was done automatically from kompose, but according to the YAML specification can be done with double, single or no quotes. The only exception is true or false as strings to avoid converting to boolean

bradbanister:
Tested out on Ubuntu with minikube, works great!

bradbanister (approved):

PR #341, opened 2024-01-17
[RND-704][RND-670] Update MetaEd and DS, begin/end dates optional on descriptors

bradbanister:


Meadowlark-js/packages/meadowlark-core/src/api-schema/ApiSchemaLoader.ts
```diff
@@ -12,7 +12,7 @@ import { ProjectNamespace } from '../model/api-schema/ProjectNamespace';
 
 const moduleName = 'ApiSchemaFinder';
 
-const dataStandard50pre1Path: string = path.resolve(__dirname, '../ds-schemas/DataStandard-5.0.0-pre.1.json');
+const dataStandard50Path: string = path.resolve(__dirname, '../ds-schemas/DataStandard-5.0.0.json');
```
stephenfuqua, line 15: I wonder if it would make sense to revert to DS 3.3 instead, since we have tests that have "3.3" in the URL, and our descriptors and the sample load process (I assume) are still from 3.3.

andonyns (dismissed):

Meadowlark-js/packages/meadowlark-core/src/api-schema/ApiSchemaLoader.ts
andonyns: This is a good point, the bulk load process has:
 $sampleDataVersion = "3.3.1-b"

Meadowlark-js/packages/meadowlark-core/src/ds-schemas/DataStandard-3.3.1-b.json
stephenfuqua, line 1: These are some big files. Which is fine at the moment. For Tanager, we'll probably want to compress the files. But we'll also need to think about the extension developer experience.

I tried to edit this file in VS code, removing all leading spaces. Wanted to see what difference that makes in the file size. VS Code completely froze when I tried to save it. Had to kill it with task manager. Notepad did better.

Result: file size from 4,2xy KB (wasn't looking closely) down to 2,501 KB.

stephenfuqua (approved):

Meadowlark-js/packages/meadowlark-core/src/ds-schemas/DataStandard-3.3.1-b.json
bradbanister: It could be trimmed down significantly. The current format has the JSON Schemas repeating three times for each resource because there are variations between insert, update and query parameters. The variations of update and query are small and can be tweaked at startup time or runtime. I pre-generated all of them in the interest of performance, but agree with the bloat. It makes it hard to inspect the file manually.

PR #363, opened 2025-03-11
Dependency updates

stephenfuqua:
There are still more vulnerabilities found by `npm audit fix`, but none could be solved without forcing the issue, and I do not have time to triage breaking changes that might occur.


stephenfuqua:
@bradbanister please review & merge. There are some unusual E2E testing errors. I would like to update these libraries now and live with the consequences of broken tests. I'll likely archive this repository after this update.

bradbanister (approved):
Yeah, not clear that the test failure is related to the dependency updates

